PR #3, opened 2020-06-11
Next lerna

im-cr:


static/readme.md
```diff
@@ -0,0 +1,33 @@
...
+    - astro.core.min.css
+    - astro.css
+    - astro.min.css
+  - icons
+    - astro.svg
+    - custom.svg
+    - status.dark.png
+    - status.light.png
+  - img
+    - astro-log-small-dark.svg
```
sknep, line 30: astro-log-small-dark.svg > astro-logo-small-dark.svg

static/readme.md
```diff
@@ -0,0 +1,33 @@
...
+
+    ` npm i astroux-static --save-dev `
+
+2. Import
+   
+    `import "astroux-static";`
+
+3. Relative Path
+   
+   `<link href="./node_modules/astro-static-css-assets/css/[Filename]>`
```
sknep, line 15: This path is incorrect. I had to use this line instead, in a style block:
`@import '/node_modules/astroux-static/css/astro.core.min.css';`

Or, with the tilde interpolation:
`@import '~astroux-static/css/astro.core.css';`

and finally, I swapped to 
`@import '/css/astro.core.css';` 
once I decided to just dump the entire static assets folder to my build directory via webpack

static/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "astroux-static",
```
sknep, line 2: I hope we namespace with with the other Astro packages in @astrouxds

Accordion has:

  "name": "@astrouxds/rux-accordion",

So maybe:

  "name": "@astrouxds/rux-static",

static/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "author": {
+    "name": "Rocket Communications UX",
+    "email": "uxsupport@rocket.com"
+  },
+  "version": "1.0.0",
+  "main": "css/astro.core.min.css",
+  "homepage": "https://www.astrouxds.com",
+  "description": "Astrouxds css and font assets for Astro UX Design System",
```
sknep, line 10: It’s nitpicky but I think we should be consistent with how we write Astro UXDS… and this has icons too, right?

`  "description": "Static CSS, SVG, and font assets for Astro UXDS",`

static/readme.md
```diff
@@ -0,0 +1,33 @@
+# Static Assets Usage
+
+## Installation
+
+1. Install the Astro UXDS Static Assets
+
+    ` npm i astroux-static --save-dev `
```
sknep, line 7: The README in the npm package that's actually published has different content on this line:
    ` npm i --save-dev astro-components-css-assets`

static/readme.md
```diff
@@ -0,0 +1,33 @@
...
+
+2. Import
+   
+    `import "astroux-static";`
+
+3. Relative Path
+   
+   `<link href="./node_modules/astro-static-css-assets/css/[Filename]>`
+
```
sknep, line 16: I don't think we can rely on the node_modules folder being available in public builds, at least not from an index.html file where someone might import this. Instead, I implemented a webpack copy that dumps the entire contents of the astro static package (or individual files) to my /dist folder:

```js
const CopyWebpackPlugin = require('copy-webpack-plugin')

module.exports = {
  configureWebpack: {
      new CopyWebpackPlugin([{
        from: 'node_modules/astroux-static/',
        toType: "dir",
        ignore: [ "index.html", "package.json", "readme.md", ".DS_Store" ]
      }])
    ]
  }
}
```
(this is the Vue way to manage webpack config changes, it uses a webpack.merge to combine with defaults appropriate for Vue)
As you an see, it would dumb the index, package.json, readme, etc files if they weren't ignored there, and it adds all 4 astro css files to the build, when only one is needed.

I tried another option where I individually imported each file, but had to set up the appropriate file loaders in webpack, which seems a lot to ask of devs new to webpack:

```js
const CopyWebpackPlugin = require('copy-webpack-plugin')

module.exports = {
  configureWebpack: {
    plugins: [
      new CopyWebpackPlugin([{
        from: 'node_modules/astroux-static/favicon.ico',
        toType: "dir",
        ignore: [ "index.html", ".DS_Store" ]
      }]),
    ]
  },
  chainWebpack: config => {
    config.module
      .rule('favicon')
      .test(/\.ico$/)
      .use('url-loader')
        .loader('url-loader')
  }
}
```

Overall, it looks like font-awesome (a similar static asset package) relies on a similar method of listing/importing assets via webpack and then assigning an appropriate loader for them:
https://medium.com/@bshelling/use-fontawesome-with-webpack-24f629d7b962

Font-awesome also has individual package variants with the specific loaders and exports for each of the major frameworks, though this seems like a lot of work:
- https://github.com/FortAwesome/vue-fontawesome
- https://github.com/FortAwesome/react-fontawesome
- https://github.com/FortAwesome/angular-fontawesome

I'm thinking at the very least, we need to include a basic webpack config in the documentation, and not expect people to know not to link to` node_modules` in production code.

static/readme.md
sknep: Also, the CSS files are all expecting the fonts to be at an absolute path, rather than a relative one, which breaks if the assets aren't always hosted at the root level of the build/dist/_site/whatever folder. In the Vue project, I changed all the font URLs in the css to relative links` url("../fonts/opensans-bold-webfont.woff2")` instead of absolute/relative to the site, as they are here: ` url("/fonts/opensans-bold-webfont.woff2")`.  Font-awesome uses a url prefixed with a variable `{$fa-font-path}` to get around this, as an alternative, but that seems messier imo.

sknep:
> Perhaps we add a how-to-use-with-webpack to the readme? 
Another, simpler option would be NCP
https://www.npmjs.com/package/ncp

PR #5, opened 2020-06-19
Add moment.JS to clock, add military timezones

sknep:
Satisfies both [#Astro-87](https://rocketcom.atlassian.net/browse/ASTRO-87) and [#Astro-143](https://rocketcom.atlassian.net/browse/ASTRO-143) by:

- Replacing date/time functions in RuxUtils used only by clock component with Moment.js package dependency, offering more accurate Day of Year count and conversion between timezones.
- Adds support for single-digit military timezone designations like "Z"

Some notes:

If we ever decide add Moment to another component (Log?) we could move the moment.js dependency (and the registration of additional military timezones) to Rux-Utils's package.json so that we aren’t importing Moment multiple times across an app. For now, though, we don't need to add Moment everywhere someone is using RuxUtils. Or maybe we leave it to webpack et al to do some tree-shaking?


src/components/rux-clock/README.md
```diff
@@ -56,7 +56,7 @@ Define AOS and LOS with valid [Unix Time Stamp](http://pubs.opengroup.org/online
...
 | `hideTimezone` | Boolean | `false` | No       | Hides the timezone in the main 24-hour clock. Timezone does not display on AOS/LOS.                                                                                                                                                                                                                     |
 | `hideDate`     | Boolean | `false` | No       | Hides the day of the year.                                                                                                                                                                                                                                                                              |
 | `small`        | Boolean | `false` | No       | Applies a smaller clock style. Previously `compact`                                                                                                                                                                                                                                                     |
```
sknep, line 62: @dmcalester I would like to list, in the revision history, the following:

- Added support for single character military designation 
- Fixed inaccurate Day of Year calculation by importing Moment.JS 

But I don't know what version this change will be released in.

src/components/rux-clock/README.md
```diff
@@ -56,7 +56,7 @@ Define AOS and LOS with valid [Unix Time Stamp](http://pubs.opengroup.org/online
...
-| `timezone`     | String  | `'UTC'` | No       | Accepts [IANA timezone string format](https://www.iana.org/time-zones) such as `America/Los_Angeles`. Default timezone is `UTC`. See [`toLocaleString()` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString#Parameters) for more details. |
+| `timezone`     | String  | `'UTC'` | No       | Accepts the [IANA timezone string format](https://www.iana.org/time-zones) such as `'America/Los_Angeles'` or any single-character designation for a [military timezones](https://en.wikipedia.org/wiki/List_of_military_time_zones) (`'A'` through `'Z'`, excluding `'J'`), both case-insensitive. Default timezone is `'UTC'`. See [`toLocaleString()` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString#Parameters) for more details.  |
```
dmcalester, line 59: Thoughts on rewording 
> Default timezone is `'UTC'`

To 
> If no timezone is defined the clock will default to UTC

stories/rux-clock.stories.js
```diff
@@ -61,9 +65,11 @@ export const ClockWithAosLos = () => {
     Alaska: 'America/Anchorage',
     Pacific: 'America/Los_Angeles',
     Mountain: 'America/Denver',
-    Central: 'America/Chicago',
+    Central: 'america/chicago',
```
dmcalester, line 68: Capitalize america and Chicago or lowercase the others?

src/components/rux-clock/rux-clock.js
```diff
@@ -1,6 +1,7 @@
 import { LitElement, html } from 'lit-element';
-// import style from './rux-clock.css';
-import RuxUtils from '../rux-utils/datetime.js';
+
+import moment from 'moment';
```
dmcalester, line 3: What is the lift to adding moment to `rux-utils`? If it’s more than 5 minutes let’s do it at another time when another component or app needs moment

dmcalester:
In thinking about this, is it really a breaking change? While the addition of moment is a new dependency an install with `npm install` should install moment and a consumer of this component shouldn’t need to change code. I would mark this as a feature update, bumping it to the next minor version over a major version, but open to counterpoints.

stories/rux-clock.stories.js
sknep: Ah, this is where i was testing that Moment.JS wasn't case-sensitive for all timezone strings. I can change it back, I just liked having it as a 'test'

src/components/rux-clock/rux-clock.js
sknep: I think this would be a bad choice to make just now, because then for all the components that load rux-utils for small stuff like the large number abbreviations (k, m,) and other formatting, they'd have this big-ish dependency they don't use. Until we have at least one more component using moment, it seems unnecessary.

stories/rux-clock.stories.js
dmcalester: No, biggie then just stood out, but the reason makes sense.

sknep:
I guess it's not really a breaking change for rux-clock, no -- anyone using the rux-clock component before just gets a slightly less buggy version now... that said, it _is_ a breaking change for rux-utils, as I removed the things rux-clock was using. Maybe I should add those back, in case someone gets a newer rux-utils on an old rux-clock?

src/components/rux-clock/README.md
dmcalester: We can release immediately, bump to 4.1.0

src/components/rux-utils/datetime.js
```diff
@@ -1,30 +1,4 @@
...
-    return new Date(
-        date.getFullYear(),
-        date.getUTCMonth(),
-        date.getUTCDate(),
-        date.getUTCHours(),
-        date.getUTCMinutes(),
-        date.getUTCSeconds()
-    );
-  }
-
```
sknep, line 27: Going to revert this -- removing these functions would be a breaking change and they don't hurt anything by staying here.

PR #6, opened 2020-06-23
Next

dmcalester:
Updating readmes and fix for Clock


src/components/rux-clock/package.json
```diff
@@ -6,7 +6,9 @@
 	"dependencies": {
 		"@astrouxds/rux-utils": "^4.0.2",
```
sknep, line 7: @dmcalester I just realized this package no longer uses rux-utils in this release

sknep: Latest push fixed this.

PR #7, opened 2020-07-25
[Astro-447] Add table styles and variables for ag-grid theme

sknep:
https://rocketcom.atlassian.net/browse/ASTRO-447

This adds a few more variables and styles necessary for a more complete ag-grid usage.

Also visible here: https://github.com/RocketCommunicationsInc/ag-grid-example/commit/054ce704415062eabe7068075560ef835bb88f66

Also see open PR for the Angular AG-grid boilerplate here: https://github.com/RocketCommunicationsInc/astro-ag-Grid/pull/3 (though this doesn't use all the new features that the above repo does)


src/css/common/_theme--light.css
```diff
@@ -7,7 +7,7 @@
   */
   --backgroundColor: var(--colorQuaternaryLighten3, rgb(243, 245, 248));
   --fontColor: var(--colorBlack, rgb(0, 0, 0));
-  --fontLowContrastColor: var(--colorGray, rgb(204, 204, 204));
+  --fontLowContrastColor: var(--colorQuaternaryDarken1, rgb(166, 171, 182));
```
sknep, line 10: The previous value of this variable wasn't used anywhere I could see, and also matched the value for the dark theme, so I assumed it was wrong. It doesn't provide nearly enough contrast for a light background use case.

This variable is used to color lower contrast items in the sidebar of the table (in both light and dark themes)

sknep:
The changes above reflect both my best efforts to more thoroughly style the ag-grid example as well as Carmen's review, indicating which colors or other attributes should be adjusted. You can follow the conversation here: https://rocketcom.atlassian.net/browse/ASTRO-449

Here's an example of the more complete theme, which uses the additional values added in this PR:
dark-ag-grid
light-ag-grid

PR #14, opened 2020-08-12
Expose selected property and change handlers on segmented button

sknep:
Satisfies https://rocketcom.atlassian.net/browse/ASTRO-552

Check the Actions tab in storybook to see the `change` event reach the parent, where the selected item is reflected to the `selected` attribute on `<rux-segmented-button>` as well as within the data array

https://deploy-preview-14--astro-components.netlify.app/?path=/story/components-segmented-button--segmented-button

### DIFF
--- a/src/components/rux-segmented-button/README.md
+++ b/src/components/rux-segmented-button/README.md
@@ -45,14 +45,14 @@ import { RuxSegmentedButton } from '@astrouxds/rux-segmented-button/rux-segmente
 
 const myButtonSegments = [
   { label: "First Segment" },
-  { label: "Second Segment", selected: true },
+  { label: "Second Segment" },
   { label: "Third Segment" }
 ];
 
 // ...
 
 render() {
-	return `<rux-segmented-button data="${myButtonSegments}"></rux-segmented-button>`;
+	return `<rux-segmented-button data="${myButtonSegments}" selected="${myButtonSegments[1].label}"></rux-segmented-button>`;
 }
 ```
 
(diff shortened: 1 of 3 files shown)


src/components/rux-segmented-button/rux-segmented-button.js
```diff
@@ -112,10 +148,10 @@ export class RuxSegmentedButton extends LitElement {
...
 
         /* 
-  OVERRIDE FOR IE 
-  Otherwise all segments get rounded corners. Need to override and re-enable
-  some style definitions.
-*/
+          OVERRIDE FOR IE 
+          Otherwise all segments get rounded corners. Need to override and re-enable
+          some style definitions.
+        */
```
dmcalester, line 155: Not important for this review, but a future ticket may exist to remove all the old `.style-scope` rules.

src/components/rux-segmented-button/rux-segmented-button.js
```diff
@@ -7,21 +7,57 @@ export class RuxSegmentedButton extends LitElement {
       data: {
         type: Array,
       },
+      selected: {
+        reflect: true,
+        type: String,
```
dmcalester, line 12: Would using the boolean property here run in to the same serialization/deserialization issues as the array and object types?

src/components/rux-segmented-button/rux-segmented-button.js
```diff
@@ -7,21 +7,57 @@ export class RuxSegmentedButton extends LitElement {
...
+    const prevSelection = this.selected;
+    this._selected = label;
+    this.data.forEach((segment) => {
+      segment.selected = segment.label === label;
+    });
+    this.requestUpdate('selected', prevSelection);
+  }
+
+  deselectSelected() {
+    const previousSelectedElement = this.querySelectorAll('input[checked]');
```
dmcalester, line 37: Could possibly use `querySelector` here instead of `querySelectorAll` since there can only ever be one selected radio button. On the other hand it’s a minor optimization vs the catch-all of All.

src/components/rux-segmented-button/rux-segmented-button.js
```diff
@@ -130,23 +166,23 @@ export class RuxSegmentedButton extends LitElement {
       </style>
 
       <ul class="rux-segmented-buttons">
-        ${this.data.map(
-      (item) => html`
+        ${this.data.map((item) => html`
```
dmcalester, line 169: Not related to this line of code, but has Storybook 6 done anything to enhance its event capturing?

src/components/rux-segmented-button/rux-segmented-button.js
sknep: I was wondering what that was about

src/components/rux-segmented-button/rux-segmented-button.js
sknep: Boolean, Number, and String types are all primitive, so they're good. Its Objects and Arrays that are problematic.

That said, a boolean can't tell you _which_ of the items has been selected, so it would need to be a string that matches a label or slug from the selected item, or a number reflecting the index of the selected item.

src/components/rux-segmented-button/rux-segmented-button.js
sknep: No, but i think I have a better handle on getting data out of the events than earlier.

sknep:
> Minor optimization question. Couple of other non-essential observations. No need for another PR.

I just realized I need to document these changes in the Readme. Stay tuned

src/components/rux-segmented-button/README.md
```diff
@@ -61,19 +61,39 @@ render() {
...
-| `selected` | Boolean | —       | No       | If true, selects this segment rather than the first segment in the `data` Array on mount. If more than one segment has a truthy `selected` value, the earliest one in the Array will register and the rest are ignored. |
+| Property   | Type    | Default | Required | Description |
+| ---------- | ------- | ------- | -------- | ------------------------------------------------------- |
+| `label`    | String  | —       | Yes      | Defines the label for the button segment. |
+| `selected` | Boolean | —       | No       | If true, selects this segment rather than the first segment in the `data` Array on mount. If more than one segment has a truthy `selected` value, the earliest one in the Array will register and the rest are ignored. Note that if the `selected` string property of the parent `rux-segmented-button` takes priority. When the selected segment changes within the component, this property updates with `true` or `false` if selected or not selected, on each segment.|
```
sknep, line 78: So technically there are two ways to pass 'selected' in -- I didn't want to break the old way to set the state, but still wanted a way to get the selected state out. In the next Major revision, I think we should steer away from passing in arrays at all, but if we don't, then I recommend selected as a string on the parent rather than a boolean on the individual children, as the later doesn't prevent someone from passing in two 'selected' children by accident, and other reasons.

It shouldn't break anyone's use cases today, but it's weird and codesmelly to have two ways to set the same thing.

src/components/rux-segmented-button/README.md
```diff
@@ -61,19 +61,39 @@ render() {
...
+
+### Events
+
+| Event Name | Description |
```
sknep, line 82: Open to thoughts about different ways to present Event documentation

dmcalester: No strong opinion. As steward of the components I defer to your guidance.

src/components/rux-segmented-button/README.md
dmcalester: What are the two ways? Setting a selected property and having the array object have a selected prop? Agree on assessment and agree selected should be explicit.

sknep:
We should release this next week, if possible

PR #19, opened 2020-09-18
ASTRO-377 - Classification Marking Banners and Labels

im-cr:
# https://rocketcom.atlassian.net/browse/ASTRO-377

WIP: Updating readme and waiting to publish `rux-classification-marking` to NPM. 

[Design Mock Reference](https://app.abstract.com/projects/2d8cf270-0b93-11e8-89ff-61554354acb1/branches/master/commits/00bfa7e635536e882e33ecc99d7311f3ec69d5dd/files/A86AC1CD-D207-4A1D-B7C0-80C938419411/layers/64234A0D-2398-44D2-84CF-799851441B26)
[Component Content Doc](https://rocketcom.atlassian.net/wiki/spaces/AS11/pages/731709441/Patterns+Classification+Control+Markings)


src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,111 @@
...
+		  color: var(--colorWhite);
+		}
+
+		:host([classification='unclassified']) {
+		  background-color: var(--classificationUnclassifiedBackgroundColor);
+		  color: var(--colorWhite);
+		}
+	  </style>
+	  
+	  <div class="rux-classification__message">${this.message}</div>
```
sknep, line 106: This allows for anyone to pass any note at all into the classifications banner and doesn't make sure that the message they typed also matches the color they chose. For example, I could pass "classified" for message and the "topsecret" type / class. I think we'd rather just have them pass one item, the classification, and have that inform the message, rather than leave this open to any input. (based on what I understand of the component requirements -- feel free to correct me if I've oversimplified the expectation)

src/components/rux-classifications/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "author": {
+    "name": "Rocket Communications",
+    "email": "uxsupport@rocketcom.com"
+  },
+  "dependencies": {
+    "@astrouxds/rux-icon": "^4.0.3",
```
sknep, line 7: Does this actually need the icons component?

src/components/rux-classifications/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "bugs": {
+    "email": "uxsupport@rocketcom.com",
+    "url": "https://github.com/RocketCommunicationsInc/astro-components/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/RocketCommunicationsInc/astro-components/src/master"
+  },
+  "description": "Astro Web Component for classification banners",
+  "license": "SEE LICENSE IN LICENSE.md",
```
sknep, line 21: Make sure to add a license.md file -- these are missing in the others on master but have been added to the v5 branch for each component folder

sknep (changes requested):
I think you're still working on the smaller labels component, but wanted to point out a few things in the banners for you.

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,111 @@
...
+      message: {
+	    type: String,
+			},
+			classification: {
+				type: String,
+			},
+			type: {
+				type: String,
+			},
+			visible: {
```
dmcalester, line 16: Check with Amalya, but I don’t think there is an instance where a classification marking should be hidden. Opening up this property could invite unintended consequences

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,111 @@
...
+				type: Boolean,
+				reflect: true,
+			},
+		};
+  }
+
+  constructor() {
+    super();
+    this.message = '';
+    this.classification = 'unclassified';
```
dmcalester, line 26: Again check with Amalya, but I suspect if there is a default we should default to the highest level of classification

src/components/rux-classifications/rux-classification.js
sknep: Here's a link to the component with the knobs set up in a way that explains what I mean: https://deploy-preview-19--astro-components.netlify.app/?path=/story/components-classification--classification-banner&knob-classification=topsecret&knob-Type=banner&knob-Classification Message=Top Secret color but I typed what I want, so UNCLASSIFIED

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,111 @@
...
+			visible: {
+				type: Boolean,
+				reflect: true,
+			},
+		};
+  }
+
+  constructor() {
+    super();
+    this.message = '';
```
dmcalester, line 25: Message shouldn't be decoupled from classification level. E.g., top secret shall always be top secret. Check with Amalya about appending a custom message after classification level, but I suspect this is unnecessary

dmcalester (changes requested):
- Please target the `next` branch not `master`
- When switching between tag and banner version the tag version fills the width of its containing element. The Classification Tags entry exhibits expected behavior of width filling to its internal content
- Validate with Amalya and Carmen the differences between font size and padding between the Sketch document, the specification and the component. The coded example appears
Screen Shot 2020-09-18 at 8 45 18 AM

src/components/rux-classifications/rux-classification.js
sknep: I wondered about this, but then thought maybe an area would conditionally have classified info... then realized you can apply `hidden` to any node if you want to hide it, so someone could do that on rux-whatever even if we didn't expose a property explicitly: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden

dmcalester: @sknep I believe with Custom Elements the `hidden` attribute doesn’t actually hide the element unless you add a specific rule to hide it. I don’t recall if we have that rule as a global property, in which case it should be considered if it should be made local to each component so that ones we absolutely don't want hidden ever like this can be made permanent. I realize that starts to toe the line of being presumptive about developer intentions.

https://developers.google.com/web/fundamentals/web-co…

src/components/rux-classifications/rux-classification.js
dmcalester: To add further clarification to this. We can retain the message/label @sknep do you have preference or guidance on label/message/other term that aligns with other components.

The message should always be appended to the broader classification marker. E.g.,
```
this.message = "Seatec Astronomy"
this.classification = "classified"
```
Would result in a component label of **CLASSIFIED//SEATEC ASTRONOMY**

The only slight outlier is TS//SCI which would be **TS//SCI//SEATEC ASTRONOMY**

src/components/rux-classifications/rux-classification.js
```diff
@@ -57,25 +57,56 @@ export class RuxClassification extends LitElement {
...
+					styleLabel = 'confidential';
+					break;
+				case 'secret':
+					styleLabel = 'secret';
+					break;
+				case 'topsecret':
+					styleLabel = 'top secret';
+					break;
+				default:
+					styleLabel = 'Top Secret//SCI';
```
sknep: @amalyah Does this make sense as the 'default' banner type (ie, most common)?

stories/rux-classifications.stories.js
```diff
@@ -15,19 +12,26 @@ export default {
...
 
   return html`
-    <div style="display: flex; flex-flow: column; justify-content: center;">
+    <div style="display: flex; flex-flow: row; justify-content: center;margin-top:60px;">
       <rux-classification-marking
         type="${markerType}"
-        classification="${classification}"
-        message="${markerText}">
+        classification="${markerFilter()}"
+        message=" ${markerText}">
```
sknep, line 34: Is this space intentional? if so, I don't think we should expect users to pass in a string that starts with a space to have the text not run together in the message.

im-cr: Yes it was intentional, but decided to remove it. As I understand it's likely that '//SCI' style abbreviations will be passed without spaces.

im-cr: Confirmed with Amalya, no space should be present.

src/components/rux-classifications/rux-classification.js
im-cr: Confirmed with Amalya, default changed back to unclassified.

src/components/rux-classifications/package.json
im-cr: added. Thanks @sknep

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,189 @@
...
+            markLabel = 'c';
+            break;
+          case 'secret':
+            markLabel = 's';
+            break;
+          case 'topsecret':
+            markLabel = 'ts';
+            break;
+          default:
+            markLabel = 'TS//SCI';
```
sknep, line 64: So you have the default class as unclassified, but the default label as TS//SCI -- they should match, no?

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,189 @@
...
+
+    return markLabel;
+  }
+
+  _setClassificationLabel(param) {
+		const markerClass = this.classification.toLowerCase().replace(/\s+/g, '');
+		const markerLabel = param.toLowerCase().replace(/\s+/g, '');
+		let markerStyle;
+		
+		if(markerLabel == markerClass){
```
sknep, line 79: What should happen if these don't match? I'm not sure what this conditional is doing for the component

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,189 @@
...
+            markLabel = 'secret';
+            break;
+          case 'topsecret':
+            markLabel = 'top secret';
+            break;
+          default:
+            markLabel = 'Top Secret//SCI';
+        }
+
+      } else if (markType === 'tag') {
```
sknep, line 45: Could you maybe combine these switch cases into one, with all the props you'd need set just once?
```js
switch(markClass) {
  case 'unclassified':
    this.markClass = 'unclassified';
    this.markerStyle = 'unclassified';
    this.markLabel = 'unclassified';
    this.tagLabel = 'u';
    break;
```

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,189 @@
...
+        type: String
+      },
+      label: {
+        type: String
+      }
+    };
+  }
+  
+  _setClassificationText() {
+    const markClass = this.classification.toLowerCase().replace(/\s+/g, '');
```
sknep, line 18: Something about having the class name be directly derived from whatever the user might pass in makes me nervous. I think we should have it select from a list of available classes, which match against the user, and provide a 'none' or default when there's no match.

Also, what happens when someone doesn't pass in a classification? What does this line return?

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,189 @@
+import { LitElement, html } from 'lit-element';
+export class RuxClassification extends LitElement {
+  static get properties() {
+    return {
+      classification: {
+        type: String
+      },
+      type: {
```
sknep, line 8: I think of "type" as String, Integer, Object, Boolean, etc. I wonder if we could be just a little more specific here in this parameter name.

Also, since there are only two supported options, what we've done in the past is choose one that's a default, and have the other be a variant  that responds to a boolean attribute present (or not present) on the element. For example:

`<rux-whatever prop="something" variantB>`

in this case, that might be a boolean attribute like "tag", which, if present, makes this not a banner.

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,189 @@
...
+      left: auto;
+			width: fit-content;
+			height: 22px;
+      padding: 4px 15px;
+      border-radius:3px;
+      font-size: var(--fontSizeMD);			
+    }
+
+    :host,
+    :host([classification='${this.getClass('topsecret//sci')}']) {
```
sknep, line 154: As we also need to maintain CSS-only versions of Astro, I think we should go ahead and hard-code (so to speak) the class names in, and make sure to add them to the Astro CSS files as well. However, I can understand wanting to keep classes consistent across CSS, HTML, and JS. You can see how we do this with the Notification banner component, which is similar.

stories/rux-classifications.stories.js
```diff
@@ -0,0 +1,183 @@
...
+		const markerClass = classification.toLowerCase();
+
+		return markerClass;
+	}
+	
+
+  return html`
+    <div style="display: flex; flex-flow: row; justify-content: center;margin-top:30px;">
+      <rux-classification-marking
+        type="${markerType}"
```
sknep, line 32: If you do change this to a Boolean attribute, then you'd want the Knob to be a checkbox, not a select.

stories/rux-classifications.stories.js
```diff
@@ -0,0 +1,183 @@
...
+    },
+  },
+};
+
+export const ClassificationBanners = () => html`
+    <div style="display: flex; flex-flow: column; justify-content: center; margin:20px;">
+      <div style="display: flex; position: relative; height: 40px; margin-bottom: 20px; overflow: hidden;">
+        <rux-classification-marking
+          type="banner"
+          classification="top secret//sci"
```
sknep, line 57: I'm not sure about accepting slashes in an HTML attribute. It doesn't look to be officially supported, even though it's working in chrome: https://www.w3.org/TR/CSS21/syndata.html#characters

Could you consider slugifying these parameters so that folks might make fewer mistakes with them? I'm imagining people who might pass it in with just one slash, or who might add spaces.

src/components/rux-classifications/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+### 3. Render the Classification Marking Component
+
+Apply `classification` property as an attribute of the Astro Classification custom element:
+
+```javascript
+<rux-classification-marking classification="controlled"></rux-classification-marking>
+
+```
```
sknep, line 47: This readme doesn't follow the format/template of the others on Astro. I think it's best to keep these consistent. It's also missing a props table and CSS-only directions. There's a template file you can use to keep with the established structure here: https://github.com/RocketCommunicationsInc/astro-components/tree/master/src/components/__rux-template

src/components/rux-classifications/rux-classification.js
```diff
@@ -0,0 +1,189 @@
...
+      flex-wrap: nowrap;
+      flex-grow: 1;		
+      width: 100%;
+    }
+    
+    :host([type='tag']){
+      position: relative;
+      align-items:center;			
+      left: auto;
+			width: fit-content;
```
sknep, line 146: I'm the last person to scold someone about whitespace, but it does look a little messy across the file here and lines 162-3, and lines 108-9.

src/components/rux-classifications/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "homepage": "https://www.astrouxds.com",
+  "bugs": {
+    "email": "uxsupport@rocketcom.com",
+    "url": "https://github.com/RocketCommunicationsInc/astro-components/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/RocketCommunicationsInc/astro-components/src/master"
+  },
+  "description": "Astro Web Component for classification banners",
```
sknep, line 19: Such a nitpick, but since these are banners and tags, maybe call it "classification markings"

src/components/rux-classifications/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+### 4. Marking Type Declaration
+ Apply the ```type``` property attribute to set the marking type. The `type` attribute property accepts two attribute values for marking generation. 1) `banner` generates a banner markingr and 2) `tag` generates a tag marking. The ```type``` attribute property value is not case sensitive.
```
sknep, line 50: "markingr" -> "marking"
"\```type```" -> "\`type`"

src/components/rux-classifications/rux-classification.js
im-cr: We can/should implement a linter config. not seeing this white space issue on my repo

src/components/rux-classifications/rux-classification.js
sknep: So when this one got changed, the other switch cases got out-of-sync

PR #21, opened 2020-09-23
Removed duplicate gray variation definition

nortonprojects:


sknep:
@dmcalester This looks like a straightforward removal -- In fact, I didn't realize we used multiple shades of Gray like this.

nortonprojects:
There's also shades of white and black FWIW

dmcalester (approved):

PR #24, opened 2020-09-28
Astro 377 - Classification Markings hotfix

im-cr:
### https://rocketcom.atlassian.net/browse/ASTRO-377

WIP: Created to address remainging updates to Classification Marking Component 

[Previous PR Notes](https://github.com/RocketCommunicationsInc/astro-components/pull/19)

### DIFF
--- a/src/components/rux-classifications/rux-classification.js
+++ b/src/components/rux-classifications/rux-classification.js
@@ -118,8 +118,7 @@ export class RuxClassification extends LitElement {
       justify-content: center;
       align-items: center;
       align-content: center; 
-      height: 26px;
-      padding: 5px 0; 
+      min-height: 26px;
       box-sizing: border-box;
       font-size: var(--fontSize);
       font-weight:bold;
@@ -127,6 +126,8 @@ export class RuxClassification extends LitElement {
       text-transform:uppercase;
       color: var(--colorBlack, rgb(0, 0, 0));
       transition: top 0.5s ease;
+      overflow-wrap: anywhere;
+      white-space: pre-line;
     }
 
     :host([type='banner']){
--- a/src/components/rux-classifications/rux-classification.js
+++ b/src/components/rux-classifications/rux-classification.js
@@ -14,59 +14,70 @@ export class RuxClassification extends LitElement {
     };
   }
   
-  _setClassificationMarking(marker) {
-    const markingClass = this.classification.toLowerCase().replace(/\s+/g, '');
-		const markingType = this.type.toLowerCase();
-		const markingLabel = marker.toLowerCase().replace(/\s+/g, '');
-		let bannerLabel;
-		let tagLabel;
-		let markingStyle;
+  _slugFilter(label) {
+    let slug = label.toLowerCase();
+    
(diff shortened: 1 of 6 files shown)


cskerbo (approved):

sknep (changes requested):
Did you try combining the switch statements into one? If it didn't work out, I'm just curious why.

Also, you have the switches' default **label** text "TOP SECRET//SCI" but the default **class** as "unclassified" -- having these in the same switch case would prevent those not matching. I can't approve this until the two defaults match, such as when someone misspells or omits an attribute. 

https://github.com/RocketCommunicationsInc/astro-components/pull/19#discussion_r494655819

src/components/rux-classifications/rux-classification.js
```diff
@@ -145,43 +114,43 @@ export class RuxClassification extends LitElement {
...
 			height: 22px;
-      padding: 4px 15px;
+      padding: 0 15px;
       border-radius:3px;
       font-size: var(--fontSizeMD);			
     }
 
     :host,
-    :host([classification='${this.getClass('topsecret//sci')}']) {
+    :host([classification='${this.marking('topsecret//sci').style}']) {
```
sknep, line 123: @im-cr How do you feel about classnames with `//` in them? It seems nonstandard to me. Did you investigate it?

src/components/rux-classifications/rux-classification.js
```diff
@@ -14,99 +14,68 @@ export class RuxClassification extends LitElement {
...
+						bannerLabel = 'secret';
+						tagLabel = 's';
+						markingStyle = 'secret';
             break;
           case 'topsecret':
-            markLabel = 'ts';
-            break;
+						bannerLabel = 'top secret';
+						tagLabel = 'ts';
+						markingStyle = 'top secret';
```
sknep, line 45: This has a space in the string, which means your class name gets broken into two? Is that intentional?

src/components/rux-classifications/rux-classification.js
dmcalester: Please remove or replace forward slashes. While valid it does introduce potential issues with escaping characters https://mathiasbynens.be/notes/css-escapes

im-cr: @sknep @dmcalester updated in last push

src/components/rux-classifications/rux-classification.js
im-cr: @sknep It is, it's used on the classification attribute so it's not two class names, just one string. I don't think this is an issue.

sknep: @im-cr would this make the class name into two classes, if you pass it in like that?

im-cr: @sknep nope, 'top secret' is treated as a single string for the classification declaration not .top.secret like a class declaration would.

The marking style output ultimately matches the human readable/writeable version of the marking classification. Like we mentioned when we discussed, it might be helpful to see if that is an issue for users. I'm open to additional thoughts.

sknep (approved):

PR #26, opened 2020-10-05
Astro 722 - Testing

nortonprojects:
This branch will be a catch-all for the different testing tasks - I'll separate each one by commit, but having 18 different pull requests seems a bit excessive.

### DIFF
--- a/__snapshots__/RuxButton.md
+++ b/__snapshots__/RuxButton.md
@@ -1,0 +1,207 @@
+# `RuxButton`
+
+#### `renders empty button correctly`
+
+```html
+<button class="rux-button">
+  <rux-icon
+    color="#ffffff"
+    hidden=""
+    icon=""
+    style="--iconDefaultColor:#ffffff;"
+  >
+  </rux-icon>
+  <slot>
+  </slot>
+</button>
+
+```
+
+#### `is default`
+
+```html
+<button class="rux-button">
+  <rux-icon
+    color="#ffffff"
+    hidden=""
+    icon=""
+    style="--iconDefaultColor:#ffffff;"
+  >
+  </rux-icon>
+  <slot>
+  </slot>
+</button>
+
+```
+
+#### `is outlined`
+
+```html
+<button class="rux-button rux-button--outline">
+  <rux-icon
+    color="rgb(0, 90, 143)"
+    hidden=""
+    icon=""
+    style="--iconDefaultColor:rgb(0, 90, 143);"
+  >
+  </rux-icon>
+  <slot>
+  </slot>
+</button>
+
+```
+
+#### `is disabled`
+
+```html
+<button
+  class="rux-button"
+  disabled=""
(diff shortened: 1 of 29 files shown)


__snapshots__/RuxButton.md
```diff
@@ -0,0 +1,207 @@
+# `RuxButton`
+
+#### `renders empty button correctly`
```
nortonprojects, line 3: The "empty button" behavior isn't documented on the Astro website, but this DOM seems reasonable. As long as it doesn't error out (and it doesn't) it should be fine.

__snapshots__/RuxButton.md
```diff
@@ -0,0 +1,207 @@
...
+    style="--iconDefaultColor:#ffffff;"
+  >
+  </rux-icon>
+  <slot>
+  </slot>
+</button>
+
+```
+
+#### `is default`
```
nortonprojects, line 20: Looks the same as above because the button text is in the slot, not the shadow-dom

src/components/rux-button/rux-button.test.js
```diff
@@ -3,14 +3,81 @@ import { html, fixture, expect } from '@open-wc/testing';
...
+  it('is large', async () => {
+    const el = await fixture(html`
+      <rux-button size="large">
+        Large Button
+      </rux-button>
+    `);
+    expect(el).shadowDom.to.equalSnapshot();
+  });
+
+  it('should ignore invalid size', async () => {
```
nortonprojects, line 67: Since the `size` property just applies a class this isn't really something to worry about at the moment, but if any other button variations are created this might become a problem (if there's a 'success' button the `size` property shouldn't be able to set that), or something like 'small large' as a value could have unintended consequences.

__snapshots__/RuxGlobalStatusBar.md
```diff
@@ -0,0 +1,77 @@
...
+#### `renders default status bar correctly`
+
+```html
+<header>
+  <div
+    class="app-meta"
+    hidden=""
+  >
+    <h1>
+      false
```
nortonprojects, line 12: I think this is just a quirk with karma rendering, it doesn't seem to show `false` in the title when actually used.

src/components/rux-icon/rux-icon.test.js
```diff
@@ -1,3 +1,78 @@
 import { html, fixture, expect } from '@open-wc/testing';
 
-import './rux-icon.js';
\ No newline at end of file
+import './rux-icon.js';
+
+describe('RuxIcon', () => {
```
nortonprojects, line 5: I think most of these should be changed to check styles or something different for better verification, but they at least make sure nothing breaks at the moment.

dmcalester:
@nortonprojects hey Jacob, do you mind if I close this PR given the forthcoming update to Stencil? cc: @cskerbo

PR #28, opened 2020-10-08
ASTRO-663 Removed moment.js in favor of date-fns

nortonprojects:


package.json
```diff
@@ -23,8 +23,11 @@
   },
   "dependencies": {
     "ag-grid-community": "^24.0.0",
+    "date-fns": "^2.16.1",
+    "date-fns-tz": "^1.0.11",
     "lit-element": "^2.3.1",
-    "lit-html": "^1.2.1"
+    "lit-html": "^1.2.1",
+    "military-timezones": "^1.0.0"
```
nortonprojects, line 30: This is basically just a json file, it would be simple to add it in to the actual code base is we didn't want to add a dependency

src/components/rux-clock/rux-clock.js
```diff
@@ -100,8 +81,17 @@ export class RuxClock extends LitElement {
...
   updateTime() {
-    this.time = moment().tz(this.timezone).format(`HH:mm:ss ${this.hideTimezone ? '' : 'z'}`);
-    this.dayOfYear = moment().tz(this.timezone).dayOfYear();
+    this.time = format(utcToZonedTime(new Date(), this._timezone), `HH:mm:ss ${this.hideTimezone ? '' : this.tzFormat}`, {timeZone: this._timezone});
+    this.dayOfYear = getDayOfYear(zonedTimeToUtc(new Date()));
+  }
+
+  convertTimezone(timezone) {
+    this._timezone = militaryTimezones[timezone.toUpperCase()];
+    this.tzFormat = 'O';
```
nortonprojects, line 90: The Intl API doesn't recognize military time zones, so this just forces it to display in terms of GMT offset when one is provided, which is what the military zones map to anyway.

package.json
dmcalester: I think it’s fine for this iteration. 2021 we’ll be looking at ways to reduce/eliminate dependencies.

cskerbo (approved):
Good work, thank you!

sknep:
What version do you think this should be released in? given that it's removing and adding a dependency, is that a Major or Minor release? It seems like the implementation itself for developers using Astro is unchanged?

nortonprojects:
I'd think it's a minor release, there shouldn't be any difference for developers

dmcalester:
A few notes. Two are essential, one is a minor note.

1. The Day of Year/Julian Date doesn’t appear to be updating. It’s Saturday in Sydney and the Julian date would be 284.
2. Non-essential but the existing clock uses the 3/4 character string value for timezones. This is not critical, timezone support is to be honest an added bonus. All DoD customers use UTC or Z
Screen Shot 2020-10-09 at 9 16 52 AM

3. The use of Z should reflect Z in the clock itself rather than GMT+0 this is a requirement from the ATLAS team
Screen Shot 2020-10-09 at 9 22 39 AM

dmcalester (changes requested):
Change request in comments

nortonprojects:
1 and 3 are fixed, but it looks like 2 is a browser thing, and it should show *most* of them using their abbreviations rather than GMT offset.

https://github.com/marnusw/date-fns-tz/issues/34

PR #29, opened 2020-10-16
Astro 403 hotfix - Component Package Name Correction

im-cr:
- Update to package naming of classification marking component.


im-cr:
A small correction to component name in component's package.json file.

PR #57, opened 2020-12-14
V2:Astro-1054 - Global Color KM Update

im-cr:
**Tickets**
  - https://rocketcom.atlassian.net/browse/ASTRO-1054

**Changes** - commit: 
  - theme__dark
  - theme__light

**Related**
  - https://rocketcom.atlassian.net/browse/ASTRO-971


im-cr:
@amalyah please confirm correct color swap for   --fontInvertedBackgroundColor in dark/light theme.css files

src/css/common/_theme--dark.css
```diff
@@ -19,12 +19,12 @@
 
 	/* styles */
   --fontColor: var(--defaultText);
-  --fontLowContrastColor: var(--colorGray);
-  --fontInvertedColor: var(--colorBlack);
-  --fontInvertedBackgroundColor: var(--colorWhite);
+  --fontLowContrastColor: var(--secondaryText);
+  --fontInvertedColor: var(--primaryElementText);
+  --fontInvertedBackgroundColor: var(--inputBackground);
```
amalyah, line 24: What are the inverted variables used for @im-cr?

src/css/common/_theme--light.css
```diff
@@ -19,12 +19,12 @@
 
   /* styles */
   --fontColor: var(--defaultText);
-  --fontLowContrastColor: var(--colorQuaternaryDarken1);
-  --fontInvertedColor: var(--colorWhite);
-  --fontInvertedBackgroundColor: var(--colorBlack);
+  --fontLowContrastColor: var(--secondaryText);
+  --fontInvertedColor: var(--primaryElementText);
+  --fontInvertedBackgroundColor: var(--primary);
```
amalyah, line 24: My suspicion is that this color will need to change, but I need to know where this color is used in order to suggest a different color.

amalyah (changes requested):

src/css/common/_theme--dark.css
im-cr: @amalyah They aren't used anywhere, just declared i the theme css files. Keeping place just in case for future proofing.

src/css/common/_theme--light.css
im-cr: Only used as declarations. Not actively being used in any css or component file.

amalyah:
I hate to delegate this out, but I think this question would be better directed at @carmenparks for visual design decision. 

> @amalyah please confirm correct color swap for --fontInvertedBackgroundColor in dark/light theme.css files

amalyah (changes requested):
Suggest asking Carmen about what colors should be used for the invertedBackground colors.  Otherwise, I suggest either using primary and the background color for both or the white/black combo for both (by using global header color awkwardly in light theme) instead of a combination of the two.

carmenparks:
> @amalyah please confirm correct color swap for --fontInvertedBackgroundColor in dark/light theme.css files

@im-cr let's go with using the primary and background colors for both themes.

amalyah (approved):
Sounds good for now since we don't have an example of what it's used in.

PR #76, opened 2021-02-08
Astro 1119 css icon implementation

Cheshire89:
https://rocketcom.atlassian.net/browse/ASTRO-1119

clean up node files and restructure to work as true cli
create script to auto generate icons css file

NOTE: this just creates css file based on available svg icons
some icons might be missing due to them being renamed or not being included in the new list
there is a separate task for displaying all the new icons based on new css implementation


dmcalester:
Hey Sasha,

In the preview deploy (https://deploy-preview-76--astro-components.netlify.app/?path=/story/components-icons-symbols--monitoring-icon-set) the `rux-icon` icons appear to be fuzzy now (and some SVG shifting) almost like a border property is being applied. Compared to the PR for ASTRO-1118 (https://deploy-preview-76--astro-components.netlify.app/?path=/story/components-icons-symbols--monitoring-icon-set) which are rendering w/o the border

cskerbo (approved):

src/css/components/icons-list.css
```diff
@@ -0,0 +1,6331 @@
+/**** DO NOT EDIT: AUTO GENERATED CSS ****/
+.rux-icon-check-box-outline-blank{
+                    -webkit-mask: url("/icons/astro.svg#check-box-outline-blank") no-repeat;
+                    mask: url("/icons/astro.svg#check-box-outline-blank") no-repeat;
+                    -webkit-mask-size: cover;
```
dmcalester, line 5: Would it make sense here to put the universally applicable properties in to the `.rux-icon` class to reduce file size.

```
.rux-icon {
 -webkit-mask-size: cover;
mask-size: cover;
-webkit-mask-repeat: no-repeat;
mask-repeat: no-repeat;
}
```

src/css/components/icons-list.css
```diff
@@ -0,0 +1,6331 @@
+/**** DO NOT EDIT: AUTO GENERATED CSS ****/
+.rux-icon-check-box-outline-blank{
+                    -webkit-mask: url("/icons/astro.svg#check-box-outline-blank") no-repeat;
```
dmcalester, line 3: I’d be interested in your thoughts on putting the urls in CSS Custom Properties. 

`--iconCheckBoxOutlineBlank: url("/icons/astro.svg#check-box-outline-blank");`

and then:

```
-webkit-mask: var(--iconCheckBoxOutlineBlank);
mask: var(--iconCheckBoxOutlineBlank);
```

I’m not sold on the idea, just curious about your thoughts. I think we’d save a small amount of file space, but I don’t know if there’s much other benefit particularly as this file is generated

src/css/components/icons-list.css
```diff
@@ -0,0 +1,6331 @@
+/**** DO NOT EDIT: AUTO GENERATED CSS ****/
+.rux-icon-check-box-outline-blank{
```
dmcalester, line 2: For the class names we generally follow a rough BEM style naming convention. Please add a double dash between `rux-icon` and its specific icon type. E.g.,

`rux-icon--check-box-outline-blank`

dmcalester (changes requested):
Hey Sasha,

I have a couple of notes. I inadvertently put them in the generated CSS file, but they may have been more appropriate in the js. If you’d like me to move them let me know.

src/css/components/icons-list.css
Cheshire89: @dmcalester  Icon list is already 6.5k lines long
creating variables and applying them will add additional 6.5k lines of code
from stand point of performance this will be an overkill

src/css/components/icons-list.css
Cheshire89: Agreed

dmcalester (approved):

PR #82, opened 2021-02-19
ASTRO-1200 toggle updates

Cheshire89:
https://rocketcom.atlassian.net/browse/ASTRO-1200

rename toggle to switch
update visual design


src/css/common/_theme--light.css
```diff
@@ -180,42 +180,18 @@
...
-
-  --toggleDisabledThumbBorderColor: var(--secondaryText);
-  --toggleDisabledThumbBackgroundColor: var(--inputBackground);
-  --toggleDisabledSelectedThumbBorderColor: var(--primary);
-
+  --switchOffColor: var(--secondaryText);
+  --switchOnColor: var(--primary);
+  --switchHoverOnColor: var(--primaryDark);
+  --switchHoverOffColor: var(--primaryDarkHover);
+  --switchDisabledOnColor: #005a8f;
```
dmcalester, line 193: If a CSS Custom property exists for this please use

dmcalester (changes requested):
When disabled the switch shows hover effect on the background track in both selected and unselected state

Hover on state seems like it might be a different color to the one in Abstract https://share.goabstract.com/d4bfbb9b-c5f2-4069-b3a5-7d11338421ef?mode=build 
Please confirm with @amalyah

src/css/common/_theme--dark.css
```diff
@@ -15,6 +15,7 @@
   --primaryLight: #92cbff;
   --primaryLightHover: #92cbff4D; /* TODO: this is a temporary fix, the use of opacity from Sketch is new and not accounted for in CSS */
 	--primaryDark: #3a87cf;
+  --primaryDarkHover: #6b8ca5;
```
amalyah, line 18: Picking the color from the design file, it looks like the opacity translates to #648EB5 over top of the Surface Element background color (where we anticipate it will be in most UIs). Please use this value instead.

src/css/common/_theme--dark.css
```diff
@@ -165,33 +166,16 @@
...
-  --toggleSelectedTrackBorderColor: var(--primaryDark);
-
-  /* Disabled State */
-  --toggleDisabledThumbBorderColor: var(--secondaryText);
-  --toggleDisabledThumbBackgroundColor: var(--inputBackground);
-  --toggleDisabledSelectedThumbBorderColor: var(--primary);
-
+  --switchOffColor: var(--secondaryText);
+  --switchOnColor: var(--primary);
+  --switchHoverOnColor: var(--primaryDark);
```
amalyah, line 175: In dark theme, the hover color should be --primaryLight.

src/css/common/_theme--light.css
```diff
@@ -180,42 +180,18 @@
...
-  --toggleSelectedTrackBackgroundColor: var(--primaryLight);
-
-  --toggleDisabledThumbBorderColor: var(--secondaryText);
-  --toggleDisabledThumbBackgroundColor: var(--inputBackground);
-  --toggleDisabledSelectedThumbBorderColor: var(--primary);
-
+  --switchOffColor: var(--secondaryText);
+  --switchOnColor: var(--primary);
+  --switchHoverOnColor: var(--primaryDark);
+  --switchHoverOffColor: var(--primaryDarkHover);
```
amalyah, line 192: Hover off color was updated in the visual designs to be 60% opacity of primaryDark on top of the Surface Element color which translates to #6691AA in our design files. There is also no separate color for the track border.

amalyah (changes requested):
- Title of branch and Jira link should change to ASTRO-1256 for accuracy
- Checked knob in Storybook isn't working.
- Dark theme: Hover state in Off position should use a different border color than the track. It's still 60% opacity of the hover color in the design files, but it's on top of the input background thumb color, so it appears lighter. This was to make the control still appear enabled.
- @dmcalester We did not design a hover state for the disabled version of the component. Do you want us to make one or can we remove it from the coded version?
- Disabled states seem to be using different colors for the track than the thumb. They should be using the same  colors as the Ready state but with the lowered opacity.

dmcalester:
> * @dmcalester We did not design a hover state for the disabled version of the component. Do you want us to make one or can we remove it from the coded version?

No. Should be removed from coded version.

src/css/common/_theme--dark.css
amalyah: This request is still relevant: Picking the color from the design file, it looks like the opacity translates to #648EB5 over top of the Surface Element background color (where we anticipate it will be in most UIs). Please use this value instead.

src/css/common/_theme--dark.css
amalyah: This request is still relevant: In dark theme, the hover color should be --primaryLight.

src/css/common/_theme--light.css
amalyah: This comment is still relevant: Hover off color was updated in the visual designs to be 60% opacity of primaryDark on top of the Surface Element color which translates to #6691AA in our design files. There is also no separate color for the track border.

amalyah (changes requested):
- Still relevant: Title of branch and Jira link should change to ASTRO-1256 for accuracy
- Still relevant: Checked knob in Storybook isn't working.
- Still relevant: Dark theme: Hover state in Off position should use a different border color than the track. It's still 60% opacity of the hover color in the design files, but it's on top of the input background thumb color, so it appears lighter. This was to make the control still appear enabled.

Cheshire89:
@amalyah 
fixed switchHoverColor in dark theme 
fixed checked knob
I will not updating the colors to Opacity version as opacity in the disabled state is being set via separate property omitting the modification of base color

cskerbo (approved):

amalyah:
> @amalyah
> fixed switchHoverColor in dark theme
> fixed checked knob
> I will not updating the colors to Opacity version as opacity in the disabled state is being set via separate property omitting the modification of base color

Re: "Still relevant: Dark theme: Hover state in Off position should use a different border color than the track. It's still 60% opacity of the hover color in the design files, but it's on top of the input background thumb color, so it appears lighter. This was to make the control still appear enabled."

@Cheshire89 Is there a technical issue/complication that stops us from making the Off Hover thumb (button) border in the dark theme appear lighter than the track color as in the Abstract file? If yes, then we can change the design instead at the point to keep the two consistent.

Cheshire89:
@amalyah 
Re: "fixed switchHoverColor in dark theme"
yes there is a technical issue ... opacity setting is literal it makes elements see through and in given case when we set opacity we get dark background coming through

Cheshire89:
@amalyah 

> > @amalyah
> > fixed switchHoverColor in dark theme
> > fixed checked knob
> > I will not updating the colors to Opacity version as opacity in the disabled state is being set via separate property omitting the modification of base color
> 
> Re: "fixed checked knob"
> 
> If you manually change the status of the switch (using the sample, not the checkmark), then the Checked knob no longer shows the correct status and doesn't change the checked status any more.
> 
> Re: "fixed switchHoverColor in dark theme"
> 
> @Cheshire89 Is there a technical issue/complication that stops us from making the Off Hover thumb (button) border in the dark theme appear lighter than the track color as in the Abstract file? If yes, then we can change the design instead at the point to keep the two consistent.

@amalyah 
The reason Off Hover color appears darker is because we are setting opacity. Which literally makes element transparent and since we have a dark back ground it makes the switch button appear darker than in design (which in turn has transparent background in abstract)

cskerbo:
@Cheshire89 Are you waiting on a re-review for this PR, or do you have additional changes?

Cheshire89:
@amalyah  @Cheshire89  Latest: 
1. Checkbox breaking the check knob 
It was a known issue in storybook  which got resolved  in future version (current 5.3 needed  5.6)
2. switchHoverOffColor was off
Fixed the color but setting opacity creates strange translucent effect on the elements of the switch button. Therefore, for now left opacity effect out

amalyah:
@Cheshire89 @dmcalester 

> 2. switchHoverOffColor was off
>    Fixed the color but setting opacity creates strange translucent effect on the elements of the switch button. Therefore, for now left opacity effect out

Is this something we will be trying to fix in the future or should I be adjusting the Hover state designs to account for this technical limitation for consistency?

src/css/common/_theme--dark.css
```diff
@@ -15,6 +15,7 @@
   --primaryLight: #92cbff;
   --primaryLightHover: #92cbff4D; /* TODO: this is a temporary fix, the use of opacity from Sketch is new and not accounted for in CSS */
```
amalyah, line 16: If we choose to continue with removing the opacity from the Switch hover states. Should we be removing this variable with the opacity value?

src/css/common/_theme--dark.css
amalyah, line 18: If we are moving forward with using the same hover color in both states of the Switch control, then do we still need this variable?

src/css/common/_theme--dark.css
```diff
@@ -158,33 +159,16 @@
...
-  --toggleSelectedTrackBorderColor: var(--primaryDark);
-
-  /* Disabled State */
-  --toggleDisabledThumbBorderColor: var(--secondaryText);
-  --toggleDisabledThumbBackgroundColor: var(--inputBackground);
-  --toggleDisabledSelectedThumbBorderColor: var(--primary);
-
+  --switchOffColor: var(--secondaryText);
+  --switchOnColor: var(--primary);
+  --switchHoverOnColor: var(--primaryLight);
```
amalyah, line 168: If we're not intending on working on the opacity issue in the near future, I suggest simplifying to a single switchHoverColor variable.

src/css/common/_theme--light.css
```diff
@@ -175,49 +175,22 @@
...
+  --switchOffColor: var(--secondaryText);
+  --switchOnColor: var(--primary);
+  --switchHoverOnColor: var(--primaryDark);
+  --switchHoverOffColor: var(--primaryDark);
+  --switchDisabledOnColor: var(--primary);
+  --switchDisabledOffColor: var(--secondaryText);
   /* 
     
-    Pushbutton Colors 
+    Popup Menu Colors
```
amalyah, line 190: Isn't the code below the pushbutton code though and not the popup menu code?

amalyah (changes requested):

src/css/common/_theme--dark.css
Cheshire89: I'll check that it is not used anywhere else. If its true it can be removed

Cheshire89: > @Cheshire89 @dmcalester
> 
> > 1. switchHoverOffColor was off
> >    Fixed the color but setting opacity creates strange translucent effect on the elements of the switch button. Therefore, for now left opacity effect out
> 
> Is this something we will be trying to fix in the future or should I be adjusting the Hover state designs to account for this technical limitation for consistency?

src/css/common/_theme--dark.css
Cheshire89: > @Cheshire89 @dmcalester
> 
> > 1. switchHoverOffColor was off
> >    Fixed the color but setting opacity creates strange translucent effect on the elements of the switch button. Therefore, for now left opacity effect out
> 
> Is this something we will be trying to fix in the future or should I be adjusting the Hover state designs to account for this technical limitation for consistency?

The issue of multiple transparent elements overlapping each other which in turn result in undesired effect will be true for any compounded element (element consisting of multiple parts). I'd suggest just picking a light value for the primary color to get similar results

Cheshire89:
ASTRO-1270: Update hover off opacity colors to solid color
https://rocketcom.atlassian.net/browse/ASTRO-1270
ASTRO-1271: Upgrade storybook to 5.6 + 
https://rocketcom.atlassian.net/browse/ASTRO-1271

src/css/common/_theme--light.css
amalyah: @Cheshire89 Still relevant: This title needs to be changed back to Push Button Colors instead of Popup Menu Colors.

amalyah (changes requested):
One last change remaining from last time

PR #92, opened 2021-03-05
Update rux-clock.js

gerhardatball:
* add tiny for a 1.0rem font-size
* add hideLabel to remove label

### DIFF
--- a/src/components/rux-clock/README.md
+++ b/src/components/rux-clock/README.md
@@ -65,7 +65,7 @@ Define AOS and LOS with valid [Unix Time Stamp](http://pubs.opengroup.org/online
 
 ## Revision History
 
-##### **4.2**
+##### **5.0**
 - Added tiny for a 1.0rem size in the clock
 - Added hideLabel to remove the label from under the field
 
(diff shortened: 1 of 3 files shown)


amalyah (approved):
If the team approves of the general concept, then I approve of the dev work. :)

cskerbo (approved):

dmcalester:
@gerhardatball thank you for your submission.

The hide label feature has been approved by our UX and dev team.

The team had a couple of questions and some possible alternatives for the **tiny** variant. 

Could you detail your particular use case for tiny? In reviewing the execution it seems the font size is the same as the small variant, but the padding around the display has been reduced to zero. Is the tiny variant intended to fit the bounding height of a particular constraint and if so what is that height? Would it be possible to share a screen shot of its use case?

Second the dev team has suggested rather than making the `isTiny` a `boolean` using the same pattern as the button component and using a `size` attribute (https://astro-components.netlify.app/?path=/story/components-buttons--standard-button) since `isTiny` and `isSmall` are mutually exclusive options. 

I want to thank you again for your submission it really meant a lot to the team

gerhardatball:
@dmcalester Thanks for the response!

I see I made a mistake in my original request. The height and font size should be reversed. That way you do still have some padding.

```
height: 1.15rem;
font-size: 1.0rem;
```
I think the suggestion of making `size` an attribute is a great idea. 

Having issues attaching screenshot right now will try again later or on Monday.

dmcalester:
> @dmcalester Thanks for the response!
> 
> I see I made a mistake in my original request. The height and font size should be reversed. That way you do still have some padding.
> 
> ```
> height: 1.15rem;
> font-size: 1.0rem;
> ```
> 
> I think the suggestion of making `size` an attribute is a great idea.
> 
> Having issues attaching screenshot right now will try again later or on Monday.

Hey Gerhard, I was wondering if you’d had any luck with screen shots? I can reach out directly via email if that would be easier.

amalyah (changes requested):
Pending further investigation from Duncan

gerhardatball:
@dmcalester Sorry I forgot to reply. I can not share a screen shot at this time. The user interface has a very minimal footer, were we are messing around with placing the clock in the footer in the bottom right without the labels.

nortonprojects (changes requested):
If this goes through next instead of master so we can track it with our usual release I'd be happy to approve

nortonprojects:
Closing due to inactivity

PR #98, opened 2021-03-15
Astro 1210 text input

Cheshire89:
https://rocketcom.atlassian.net/browse/ASTRO-1210
light styling update for text inputs


dmcalester (approved):

cskerbo (approved):

src/css/core/input.css
```diff
@@ -84,12 +85,25 @@
   border-radius: 3px;
```
Design-Kat, line 85: Unless there is a separate ticket for text-areas, the border and font changes should be reflected in text area as well. I'm still verifying the border- it appears it was changed and I want to confirm as I believe it was originally designed to match the buttons, will update when I've got the confirmation.

Design-Kat: @Cheshire89 Confirmed with Ty that it actually should be 3px border radius- I will put in a change for the sketch file for this.

src/css/core/input.css
```diff
@@ -60,12 +61,12 @@
 
   height: 2.125rem;
   width: 100%;
-  padding: 0 1rem;
+  padding: 0 0.625rem;
 
   border: 1px solid var(--inputBorderColor);
-  border-radius: 3px;
+  border-radius: 4px;
```
Design-Kat, line 67: @Cheshire89 Confirmed with Ty that it actually should be 3px border radius- I will put in a change for the sketch file for this.

static/css/astro.core.css
```diff
@@ -2022,14 +2023,14 @@ label {
 
   height: 2.125rem;
   width: 100%;
-  padding: 0 1rem;
+  padding: 0 0.625rem;
 
   border: 1px solid var(--inputBorderColor);
-  border-radius: 3px;
+  border-radius: 4px;
```
Design-Kat, line 2029: @Cheshire89 Confirmed with Ty that it actually should be 3px border radius- I will put in a change for the sketch file for this.

Design-Kat (changes requested):
So, 2 things: 1) It seems that the sketch file has the input's border radius changed to 4px, I have confirmed with Ty that it should match the 3px border radius of buttons. 2) Any changes made to text inputs should probably matched on Text areas (I can't think of a reason why they wouldn't be, @amalyah @dmcalester could you confirm please? There's no special reason I'm not thinking of, is there?)

Design-Kat (approved):

PR #99, opened 2021-03-15
Astro 1208 number inputs

Cheshire89:
https://rocketcom.atlassian.net/browse/ASTRO-1208
Show native styling for number inputs


dmcalester (changes requested):
Couple of quick changes.

- Change placeholder text to sentence case
- The invalid state with the number spinners is covered. We’re run in to this before (which is where hiding the spinners came about as an interim solution that lasted too long)
Screen Shot 2021-03-15 at 1 23 37 PM

amalyah:
> Couple of quick changes.
> 
> * Change placeholder text to sentence case
> * The invalid state with the number spinners is covered. We’re run in to this before (which is where hiding the spinners came about as an interim solution that lasted too long)
> 
> Screen Shot 2021-03-15 at 1 23 37 PM

FYI it might not be worth fixing this since the invalid icon will be moving out of the inputs in another ticket.

cskerbo (approved):

Design-Kat (changes requested):
Not sure if this would be handled with this ticket, or a bug: The middle column of the Smaller section has some odd layout issues happening (it is not aligned to the top of the grouping like column 1 and 3 are) - Not sure if that's due to how the fields are laid out/wrap/etc or if it's a genuine bug, but thought I'd mention it as a possible issue. It only seems to align correctly when the page is smaller and text labels start wrapping (which those off the normal text field's middle column). 
Other than that I'm not seeing any issues.

dmcalester (approved):

Cheshire89:
> Couple of quick changes.
> 
> * Change placeholder text to sentence case
> 
> I withdraw this comment the below
> 
> * The invalid state with the number spinners is covered. We’re run in to this before (which is where hiding the spinners came about as an interim solution that lasted too long)
> 
> Screen Shot 2021-03-15 at 1 23 37 PM

I'll just move the position of error icon if its a number input

Cheshire89:
@Design-Kat  i've added media queries to change column count based on window size. This should help with labels wrapping 
As for error icon, please let me know if it needs to be fixed. I've let it be for now since it will be fixed in later task

Design-Kat:
> @Design-Kat i've added media queries to change column count based on window size. This should help with labels wrapping
> As for error icon, please let me know if it needs to be fixed. I've let it be for now since it will be fixed in later task

@Cheshire89 Sorry for the delay, I missed this notification - I think that's a @dmcalester call (on whether the error icon should be fixed now vs later, when it will be fixed/moved out of the input by the Validation tickets). Personally if it's going to be there for a month or more this way, I lean towards fixing, but if it's going to be there a week, it's not worth it, but that's not my call to make.

Design-Kat:
> Not sure if this would be handled with this ticket, or a bug: The middle column of the Smaller section has some odd layout issues happening (it is not aligned to the top of the grouping like column 1 and 3 are) - Not sure if that's due to how the fields are laid out/wrap/etc or if it's a genuine bug, but thought I'd mention it as a possible issue. It only seems to align correctly when the page is smaller and text labels start wrapping (which those off the normal text field's middle column).
> Other than that I'm not seeing any issues other than what's already been said.

It looks much better- I still do see the 2nd column of the "smaller" text inputs slightly lower, but I suspect it's a Safari-only issue. Chatted w/ Amalya about it and she mentioned the layout will likely need to change for Validation so it's likely not an issue for this ticket, but I wanted to make you aware that there is still an odd dip that you may want to talk to @dmcalester to see about whether Safari is a fully supported browser or not.

Design-Kat (approved):
Left some comments but I don't think they're really "number input" issues, so not really a bar to approval.

PR #101, opened 2021-03-17
ASTRO-1275

nortonprojects:
https://rocketcom.atlassian.net/browse/ASTRO-1275


src/components/rux-button/rux-button.js
```diff
@@ -17,6 +17,13 @@ export class RuxButton extends LitElement {
     (this.icon = ''), (this.iconOnly = false);
     this.disabled = false;
     this.outline = false;
+    this.addEventListener('click', this.clickHandler);
+  }
+
+  clickHandler(e) {
+    if (this.disabled) {
+      e.stopImmediatePropagation();
```
dmcalester, line 25: @nortonprojects a couple of quick questions, mostly just so I understand.

- Is this something we should add to all components with click events that can be disabled?
- Is there a difference between this and `preventDefault()`?
- Should we also account for keyboard/ENTER events?

dmcalester (approved):

cskerbo (approved):

src/components/rux-button/rux-button.js
nortonprojects: https://codeplanetio.wordpress.com/2015/05/19/preventdefault-vs-stoppropagation-vs-stopimmediatepropagation/#:~:text=preventDefault%3A%20Cancels%20the%20event%20if,same%20event%20from%20being%20called.

Basically, stopImmediatePropogation() removes the event entirely, while preventDefault() just cancels the normal browser actions.

And yes, we should add this to the other components as well

PR #100, opened 2021-03-17
ASTRO-1207

nortonprojects:
https://rocketcom.atlassian.net/browse/ASTRO-1207

PR #103, opened 2021-03-18
Astro 1111 login component

Cheshire89:
https://rocketcom.atlassian.net/browse/ASTRO-1111
Basic login component with main functionality
DId best I could with designs ... considering they keep on changing


src/components/rux-login/rux-login.js
```diff
@@ -0,0 +1,333 @@
...
+                  <rux-icon icon="lock" size="extra-small" color="${this.sso ? 'var(--primary)':'var(--secondaryText)'}"></rux-icon>
+                  SSO Enabled
+                </p>
+              </div>
+            </div>
+
+            ${!this.sso ? html`
+              <div class="rux-form__group">
+                <div class="rux-form-field rux-form-field--withIcon">
+                  <label for="pass">Password ${this.passwordVisible}</label>
```
cskerbo, line 302: Should `${this.passwordVisible}` actually be inside the label?

Screen Shot 2021-03-18 at 3 19 13 PM

cskerbo (changes requested):
When typing in either of the input fields, the login component forces fullscreen in the Storybook - is this intentional?

It also seemed to bounce back and forth between default and fullscreen on each key press.

amalyah:
FYI @Design-Kat this is up for review.

src/components/rux-login/rux-login.js
Cheshire89: Oops ... as for the screen view changing ... I've had it once but thought I've pressed something instead of it being a bug

Cheshire89:
@tyvanleuven you're not in the list may you'll get a memo this way ? Need your review here as well

src/components/rux-login/rux-login.js
Design-Kat: I am still seeing this issue in Safari, btw.

src/components/rux-login/rux-login.js
```diff
@@ -0,0 +1,332 @@
...
+            }
+        </style>
+
+        <form name="ruxLogin" id="ruxLogin" class="rux-form">
+            <div class="rux-form__group">
+              <div class="rux-form-field">
+                <label for="username">Username</label>
+                <input 
+                  type="email"
+                  id="username"
```
Design-Kat, line 283: To match our documentation, we should use "Email" in the UI for display, rather than Username.

src/components/rux-login/rux-login.js
```diff
@@ -0,0 +1,332 @@
...
+                  class="rux-input"
+                  .value="${this.username}"
+                  @input="${this._updateValue}"
+                  placeholder="Email@astro.com"
+                />
+              </div>
+              <div class="rux-form-field">
+                <p class="rux-form__help" style="color: ${this.sso? 'var(--primary)': ''}" @click="${this._changeSSO}">
+                  <rux-icon icon="lock" size="extra-small" color="${this.sso ? 'var(--primary)':'var(--secondaryText)'}"></rux-icon>
+                  SSO Enabled
```
Design-Kat, line 293: In Safari I am seeing the "SSO Enabled" w/ lock underneath the login, that line should only display when the login is being designed for SSO entry. The Lock icon should also display as Primary Blue, rather than white.

src/components/rux-login/rux-login.js
```diff
@@ -0,0 +1,332 @@
...
+
+            ${!this.sso ? html`
+              <div class="rux-form__group">
+                <div class="rux-form-field rux-form-field--withIcon">
+                  <label for="pass">Password ${this.passwordVisible}</label>
+                  <input
+                    type="${this.passwordVisible ? 'text' : 'password'}"
+                    id="pass"
+                    class="rux-input"  />
+                  <rux-icon
```
Design-Kat, line 306: I'm seeing this icon display as 15px wide, the design has it a bit bigger (so that it fits in a square box that is 24px wide).

Design-Kat (changes requested):
Added a few comments. Please let me know if you have any questions! (Ty can also speak to the visuals.)

src/components/rux-login/README.md
```diff
@@ -0,0 +1,3 @@
+# Login
```
amalyah, line 1: I believe this component was actually called Sign In in the end. Would need to change the various file titles and function names as well for proper reference. Has that been changed @Design-Kat ?

src/components/rux-login/README.md
```diff
@@ -0,0 +1,3 @@
...
+
+Login enables a user to pass a user name and a password to the backend for authentication verification
```
amalyah, line 3: Login > Sign In

verification > verification.

src/components/rux-login/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "homepage": "https://www.astrouxds.com",
+  "bugs": {
+    "email": "uxsupport@rocketcom.com",
+    "url": "https://github.com/RocketCommunicationsInc/astro-components/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/RocketCommunicationsInc/astro-components/src/master"
+  },
+  "description": "Astro Web Component for login",
```
amalyah, line 21: login > Sign In

src/components/rux-login/rux-login.js
```diff
@@ -0,0 +1,332 @@
+import { LitElement, html, css } from 'lit-element';
+import { RuxIcon } from '@astrouxds/rux-icon';
+import { RuxButton } from '@astrouxds/rux-button';
+
+export class RuxLogin extends LitElement {
+  static get properties() {
+    return {
+      username: {
```
amalyah, line 8: According to the visual design, this should be Email instead of Username.

amalyah (changes requested):
- My largest concern is that I think this is supposed to be called Sign In instead of Login across Astro.
- @Design-Kat Is the SSO enabled part supposed to be a link? I thought they were two different examples altogether.
- SSO text and its icon is 1 rem in the design.
- I agree with the previous comment that the false/true part next to the Password label should be removed for publishing.

src/components/rux-login/README.md
Design-Kat: Good catch, you are correct, unless it was discussed with @dmcalester and there's need for a separate distinction for some circumstances?

Cheshire89:
@amalyah && @Design-Kat  
renamed everything to from login to sign in
removed true / false test code
updated username field to email

src/components/rux-sign-in/rux-sign-in.js
```diff
@@ -0,0 +1,332 @@
...
+            .ml-auto{
+              margin-left: auto;
+            }
+
+            .rux-form-field > label {
+              margin-bottom: 0.5rem
+            }
+        </style>
+
+        <form name="ruxLogin" id="ruxLogin" class="rux-form">
```
Design-Kat, line 277: Should these instances be Signin- or is it only the items visible to the user on the website that were changed?

src/components/rux-sign-in/README.md
```diff
@@ -0,0 +1,3 @@
+# Sign in
+
+Sign in compoennt enables a user to pass an email and a password to the backend for authentication verification
```
Design-Kat, line 3: "compoennt" should be "component"

src/components/rux-sign-in/rux-sign-in.js
dmcalester: This is ok

Cheshire89:
image

I dot nline the copy of title here

Design-Kat (changes requested):
Seeing a spelling mistake (compoennt > component) on readme. 
Lock and Visibility icon is still reading as  smaller than they should be. Ty will need to double check for approval of those.
 Still seeing “SSO Enabled” under the email field- that is only supposed to be visible in SSO mode. Seeing this in both safari and chrome.

amalyah (changes requested):
SSO Enabled text and icon still causes the cursor to change to a hand which makes it look interactable, which it is not.

src/components/rux-sign-in/rux-sign-in.js
```diff
@@ -0,0 +1,327 @@
...
+              font-size: 16px;
+              color: var(--controlLabelColor);
+              display: flex;
+              align-items: center;
+              margin: 0px;
+              margin-top: 8px;
+            }
+
+            .rux-form__help:hover{
+              cursor: pointer;
```
cskerbo, line 241: @Cheshire89 I am still seeing the "SSO Enabled" causing a cursor change. Is it because it's nested in the form and taking this style, perhaps?

You may have to apply direct styling of `cursor: none` instead of removing the pointer as your previously did.

cskerbo (changes requested):

Cheshire89:
So sorry for that SSO thing :(

dmcalester (approved):
Approved with a lien for after PI to add README documentation https://rocketcom.atlassian.net/browse/ASTRO-1291

amalyah (changes requested):
Shoot. Just realized that the email field is using the wrong invalid styling. The error icon is still inside the field instead of providing error text below the field with the icon.

Cheshire89:
> Shoot. Just realized that the email field is using the wrong invalid styling. The error icon is still inside the field instead of providing error text below the field with the icon.

Isn't there a separate card for this ?  Because that will involve all the input fields

amalyah:
> > Shoot. Just realized that the email field is using the wrong invalid styling. The error icon is still inside the field instead of providing error text below the field with the icon.
> 
> @amalyah Isn't there a separate card for this ? Because that will involve all the input fields

You are totally right, @Cheshire89. Looks like that fix was supposed to be completed in [PR 100](https://github.com/RocketCommunicationsInc/astro-components/pull/100) by @nortonprojects. @nortonprojects the problem that I mentioned in that PR still seems to be re-occuring. Could you please look into this?

Design-Kat:
The only issue I'm seeing are the issues that Amalya had already brought up with the Invalid state, but I don't think it needs a disapproval when they're already mentioned. So once those issues have been sorted (this ticket/another?) I'm good with this implementation.

Cheshire89:
> The only issue I'm seeing are the issues that Amalya had already brought up with the Invalid state, but I don't think it needs a disapproval when they're already mentioned. So once those issues have been sorted (this ticket/another?) I'm good with this implementation.

What is exactly wrong with invalid state. Error messages is exactly what I've added

cskerbo (approved):

Design-Kat (approved):
Approving with the added note that the validation summary (Email or password not found, yet not "tied" to a specific field) doesn't have official guidance, and may eventually need a ticket.

@dmcalester do you want another ticket split out for server-side type errors? (ie: access denied/lockout), or are you comfortable with the guidance being messaging in a modal dialog per the sign in documentation?

Cheshire89:
> Approving with the added note that the validation summary (Email or password not found, yet not "tied" to a specific field) doesn't have official guidance, and may eventually need a ticket.
> 
> @dmcalester do you want another ticket split out for server-side type errors? (ie: access denied/lockout), or are you comfortable with the guidance being messaging in a modal dialog per the sign in documentation?

It is tied to invalid property of component which can be accessed

PR #106, opened 2021-03-25
Astro 1288 - Scrub through components to apply disabled fix

markacianfrani:
https://rocketcom.atlassian.net/browse/ASTRO-1288

Button, PushButton, Slider, Switch, and Tab look like the only components with an explicit disabled prop. Let me know if I may have missed any other places.


dmcalester (approved):

cskerbo (approved):
I believe you caught them all, thank you!

PR #108, opened 2021-03-30
Astro 1328

markacianfrani:
The issue here was that ```_registerTabPanels()``` wasn't firing when the component was being used in Storybook because the component is listening for the DOMContentLoaded event. I suspect this was already fired once somewhere in Storybook's lifecycle. To solve this, we can [check if loading is already complete](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete) and execute ```_registerTabPanels()``` 

I don't know the historical context of why that event listener exists but it looks like it's related to an Angular issue. I tried to avoid making any changes to the component just to fix an issue with Storybook specifically, but I couldn't come up with any other solutions.

### DIFF
--- a/src/components/rux-tabs/rux-tab-panels.js
+++ b/src/components/rux-tabs/rux-tab-panels.js
@@ -16,7 +16,7 @@ export class RuxTabPanels extends LitElement {
       window.addEventListener('DOMContentLoaded', this._registerTabPanelsListener);
     } else {
       // Register Tab Panels if DOMContentLoaded event was already fired
-      this._registerTabPanelsListener;
+      this._registerTabPanelsListener();
     }
   }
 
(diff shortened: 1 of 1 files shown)


dmcalester:
> I don't know the historical context of why that event listener exists but it looks like it's related to an Angular issue. I tried to avoid making any changes to the component just to fix an issue with Storybook specifically, but I couldn't come up with any other solutions.

Hi Mark, that sounds exactly right. Do you know if this change would break its use in Angular and if so are there any steps we can take to support both?

markacianfrani:
@dmcalester I'll fire up a fresh Angular project to be safe, but I can't think of a reason why it would break. It should fall back to listening for DOMContentLoaded

markacianfrani:
It actually does break in Angular, unless you wrap the component in a conditional like ngIf. Back to the drawing board!

PR #112, opened 2021-04-01
ASTRO-1087 - fix: height and alignment on classification portion markers

markacianfrani:
fixes height to 22px and aligns text to match source document


Full-lifey (commented):
Good to see relative units.

@Cheshire89 Does this css need to be copied? That might be due to the old gulp setup but it's worth asking.

Cheshire89:
> Good to see relative units.
> 
> @Cheshire89 Does this css need to be copied? That might be due to the old gulp setup but it's worth asking.

Since its in the js file only it will have no effect upon what i did

Cheshire89 (approved):

cskerbo (approved):

PR #113, opened 2021-04-02
Astro 445 

markacianfrani:
Fixes an issue with the notification component not closing entirely and throwing an error when using the closeAfter feature. The error was related to trying to assign a value to `_closeAfter`, but there was no setter. 

I'm not sure if clearTimeout is even necessary here though. I'm assuming thats there to prevent any potential memory leaks?

### DIFF
--- a/src/components/rux-notification/rux-notification.js
+++ b/src/components/rux-notification/rux-notification.js
@@ -74,7 +74,7 @@ export class RuxNotification extends LitElement {
           align-items: center;
           align-content: center;
 
-          top: -4.25rem;
+          top: -4.375rem;
           left: 0;
 
           height: 4.375rem;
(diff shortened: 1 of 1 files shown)


src/components/rux-notification/rux-notification.js
```diff
@@ -38,25 +38,27 @@ export class RuxNotification extends LitElement {
 
   updated() {
     if (this._closeAfter && this.open) {
-      this._closeAfter = setTimeout(() => {
+      this.timerId = setTimeout(() => {
```
Cheshire89, line 41: There is no initializer in the constructor for the timeId which may cause errors in some cases. And I'd call time timeoutRef to be more intuitive

Cheshire89 (changes requested):
small tweak

src/components/rux-notification/rux-notification.js
markacianfrani: @Cheshire89 good idea, thanks!

cskerbo (approved):

PR #114, opened 2021-04-05
README typo fix

flash548:
Typo fix for stylesheet import


dmcalester (approved):
Thank you for the catch.

PR #116, opened 2021-04-15
ASTRO-1319 extend pop up menu to allow for custom content

Cheshire89:
https://rocketcom.atlassian.net/browse/ASTRO-1336

Allow insertion of custom content into pop up menu

I've tried a mirriad of things to make the custom event binding happen but we are limited by winodow scoped events that LitElement uses.

### DIFF
--- a/src/components/rux-pop-up-menu/rux-pop-up-menu.js
+++ b/src/components/rux-pop-up-menu/rux-pop-up-menu.js
@@ -16,6 +16,12 @@ export class RuxPopUpMenu extends LitElement {
       _trigger: {
         type: Object,
       },
+      popUpMenuItemSelected: {
+        type: Function
+      },
+      popUpMenuExpandedChange: {
+        type: Function
+      }
     };
   }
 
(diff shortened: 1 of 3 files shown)


src/components/rux-pop-up-menu/rux-pop-up-menu.js
```diff
@@ -172,21 +155,23 @@ export class RuxPopUpMenu extends LitElement {
   }
 
   render() {
+    const list = this.data.map((item, index) => {
+      return item.rule === 'separator' ?
```
markacianfrani, line 159: Looks like a typo here. `return item.role === 'seperator' ?`

dmcalester (changes requested):
A couple of notes

1. The separator element behaves like an active element. I am able to click on it and it highlights. This should be an inert element.
2. The separator element is not displaying the dotted line of the published iteration
3. Clicking on a menu item does not close the popup menu (was this requested behavior; we might need to push back on this)
4. The menu doesn’t close when using a "clickOutside" event (again was this a request? Might need to push back or add variant :grimace:
5. Related to #4 clicking on one menu and then another the first menu stays open
6. Possibly explaining 3-5 getting a JS error when clicking on a menu item
`this.onPopUpMenuExpandedChange is not a function`

cskerbo (approved):

markacianfrani (approved):
Should we change the long item text back to "Item 5 with a Really Long Title ..." as per the EGS compliance shown here https://www.astrouxds.com/components/pop-up/ ? Or is that outdated? Looks good to me!

dmcalester (approved):

PR #119, opened 2021-04-19
Astro 1353 icons update

Cheshire89:
https://rocketcom.atlassian.net/browse/ASTRO-1353

update node commands to generate icons json file which includes style and category of each icon
update icons page to include icon category
fix grid styling

remove following icons 

atm
beach-access
cake
local-car-wash
local-florist
local-pizza
polymer
room-service


dmcalester (approved):
Hey Sasha, no biggie, but would it be possible to have the last row of icons use the the equivalent of `flex-start` so each icon aligns under the one above it?

So like this
```
[] [] [] [] []
[] []
```

Instead of this
```
[] [] [] [] []
    []   []
```

PR #121, opened 2021-04-23
fix sign in form missing styles

Cheshire89:
styling from the fom was missing due to new injection


dmcalester (approved):
Hey Sasha, what are the !! and ! 

Just curious.

PR #124, opened 2021-05-04
fix: change font weight on aos/los clock

markacianfrani:
fixing very minor inconsistency in rux-clock where the aos/los labels font weight is defined as 100 but the design calls for 400. because we don't include Roboto Mono 100, its being rendered as 400 anyway so this effectively changes absolutely nothing visually.


dmcalester (approved):
Thanks Mark

PR #125, opened 2021-05-05
Hot fix for erroneous colors

dmcalester:
NOTE: I believe some of the color discrepencies reported by L3/Harris
may be coming from the conversion where the Gulp/CSSNext generator
created the stepped color increments.

This is a triage fix to addres the basic concerns. This should be
addressed in a more comprehensive design token effort


cskerbo (approved):

markacianfrani (approved):
@dmcalester should this be merged into next as well?

PR #130, opened 2021-05-26
update typograph sizes per spec

Cheshire89:
typography edits didnt go through for some reason
https://rocketcom.atlassian.net/browse/ASTRO-1444?atlOrigin=eyJpIjoiZWUyYmQwNjFjZDI2NDIwY2I0OTg4YTNhZGFmYmM1MWIiLCJwIjoiaiJ9


Cheshire89:
@avillaindesire your not in the reviewers list so i'm trying to ping you this way

dmcalester (approved):

src/scss/common/_typography.scss
```diff
@@ -2,24 +2,25 @@ h1,
 h2,
 h3 {
 	font-family: var(--fontFamily);
-  font-weight: 300;
+  font-weight: 400;
```
amalyah, line 5: Suggest removing non-shared font-weight (400) from this shared declaration.

static/css/astro.core.css
```diff
@@ -648,19 +648,20 @@ li.light-theme {
...
   --primaryLightHover: #92cbff4D;
   /* TODO: this is a temporary fix, the use of opacity from Sketch is new and not accounted for in CSS */
-  --primaryDark: #3a87cf;
+  --primaryDark: var(--colorSecondaryLighten2, #3a87cf);
   --primaryDarkHover: #6b8ca5;
-  --primaryElementText: #080c11;
-  --inputBackground: #FFFFFF;
+  --primaryElementText: var(--colorTertiaryDarken4, #080c11);
+  --inputBackground: var(--colorWhite, #FFFFFF);
   --inputDark: #080c11;
```
amalyah, line 665: This should also be --colorTertiaryDarken4.

static/css/astro.core.css
```diff
@@ -1023,21 +1024,22 @@ h2,
 h3 {
   font-family: "Roboto", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
   font-family: var(--fontFamily);
-  font-weight: 300;
+  font-weight: 400;
```
amalyah, line 1027: Suggest removing non-shared font-weight (400) from this shared declaration.

static/css/astro.core.css
```diff
@@ -648,19 +648,20 @@ li.light-theme {
...
   --surfaceElements: #1b2d3e;
-  --primary: #4dacff;
-  --primaryLight: #92cbff;
+  /* TODO: this is an unofficial Astro color, but a required KM color */
+  --primary: var(--colorSecondary, #4dacff);
+  --primaryLight: var(--colorSecondaryLighten2, #92cbff);
   --primaryLightHover: #92cbff4D;
   /* TODO: this is a temporary fix, the use of opacity from Sketch is new and not accounted for in CSS */
-  --primaryDark: #3a87cf;
+  --primaryDark: var(--colorSecondaryLighten2, #3a87cf);
```
amalyah, line 661: This should be --colorSecondaryDarken1

static/css/astro.core.css
dmcalester: Compiled CSS

src/scss/common/_typography.scss
Cheshire89: It is being over written for the h3 tag lower in the code
however h1 and h2 are 400 weight

amalyah: Is it worth having it here and not just in the individual heading definitions?

Cheshire89: yes it is accepted css practice

static/css/astro.core.css
Cheshire89: computed css files should not be reviewed i.e anything that starts with `astro`

 As they are triggered by compiler which might be affected many different things such as code formatters, variable formatters (which generate our RGB values) for obvious reasons we either have limited control over output or tweaks will lead to global changes

Full-lifey (approved):

dmcalester (approved):

PR #146, opened 2021-06-17
Adds prettier config

NicDominguez:
Updates Prettier to same version as the stencil repo and installs necessary tools to run prettier as a pre-commit hook. All non-default config options are set to the same as the stencil repo. Config file and husky directory were whitelisted in the gitignore file to insure they remain consistent when each of us work on this repo.

Reminder: you will need to run npm install after you pull after this merge to make sure you have the necessary tools installed.


markacianfrani (changes requested):
might want to add in the npm prettier command to make it easier to run the initial pass. maybe not because it'll probably only be ran once. 

can you add the prettierignore file though? (it was added while you were starting this I think). You'll need to replace some of the directories though. we want to exclude any build artifacts (probably storybook static?) and node_modules

markacianfrani (approved):

micahjones13 (approved):

PR #156, opened 2021-07-08
Astro 1556 scroll bar km

micahjones13:
Updates scroll bar styles to be inline with current abstract implementation and KM design changes.

[Jira Ticket](https://rocketcom.atlassian.net/browse/ASTRO-1556)

Dark Theme changes: 

1. Changed `--scrollBarThumbBackgroundColor:` to `rgb(43, 101, 155)` from `rgb(46, 103, 153)`
2. Changed `scrollBarTrackCornerBackgroundColor` to  `--scrollBarTrackBackgroundColor` and changed color to `rgb(27, 45, 62)` from `rgb(32, 50, 70)`
3. Created `scrollBarVerticalBoxShadow` and `scrollBarHorizontalBoxShadow` with values `inset 3px 3px 3px 0px rgba(0, 0, 0, 0.5);` and `inset 1px 3px 3px 0px rgba(0, 0, 0, 0.5)` respectively.

Light Theme changes: 

1. changed `--scrollBarThumbBackgroundColor` to `var(--primary, rgb(0, 90, 143))` from `var(--colorQuaternaryDarken1, #9ba0aa)`
2. Changed `scrollBarTrackCornerBackgroundColor` to  `--scrollBarTrackBackgroundColor` and changed color to `var(--colorWhite, rgb(255, 255, 255));` from `transparent`
3.  Created `scrollBarVerticalBoxShadow` and `scrollBarHorizontalBoxShadow` with values `inset 3px 3px 3px 0px rgba(47, 72, 99, 0.5);` and `inset 1px 3px 3px 0px rgba(47, 72, 99, 0.5)` respectively.

Some values are not using pre-existing colors (ie. rgb(43, 101, 155) and rgb(27, 45, 62)) and so I left them as is. We will most likely be auditing the color values and var names in the future, and these can be updated along with that audit.


src/scss/common/_global.scss
```diff
@@ -43,41 +44,42 @@ label {
...
 }
 
 ::-webkit-scrollbar-thumb {
-  background-color: var(--scrollBarThumbBackgroundColor, rgb(46, 103, 153));
-  border-radius: 10px;
-  border: 3px solid transparent;
-  background-clip: padding-box;
+    background-color: var(--scrollBarThumbBackgroundColor, rgb(43, 101, 155));
+    border-radius: 10px;
+    // border-radius: 4px;
```
micahjones13, line 55: Abstract has border-radius at 4px, but it doesn't look anything like abstract's implementation if 4 is provided here.

markacianfrani (changes requested):
for some reason, it looks like this breaks the select form component styles. can you take a look? its hard for me to see what's going on in the diff history

src/scss/common/_global.scss
amalyah: In Sketch we use 4px for the radius of the thumb.

src/scss/common/_global.scss
```diff
@@ -43,41 +44,38 @@ label {
...
 
 ::-webkit-scrollbar-thumb {
-  background-color: var(--scrollBarThumbBackgroundColor, rgb(46, 103, 153));
-  border-radius: 10px;
-  border: 3px solid transparent;
-  background-clip: padding-box;
+    background-color: var(--scrollBarThumbBackgroundColor);
+    border-radius: 10px;
+    // border-radius: 4px;
+    border: 3px solid transparent;
```
amalyah, line 56: Is there a reason to have a transparent border? Just a placeholder for if we ever add a border later?

src/scss/common/_global.scss
```diff
@@ -43,41 +44,38 @@ label {
...
+    border-top-width: 4px;
 }
 
 ::-webkit-scrollbar-thumb:active,
 ::-webkit-scrollbar-thumb:hover {
-  background-color: var(
-    --scrollBarThumbBackgroundHoverColor,
-    rgb(58, 129, 191)
-  );
+    background-color: var(--scrollBarThumbBackgroundHoverColor);
```
amalyah, line 70: Having a different color when the thumb is active or hovered is not recorded in the design files. @avillaindesire is this something you consulted with dev on? (If yes, we should add those states to the design files. If no, we should remove it from code.)

src/scss/common/_theme--dark.scss
```diff
@@ -354,23 +354,14 @@
...
   */
-
-    --scrollBarThumbBackgroundColor: var(
-        --colorSecondaryDarken2,
-        rgb(46, 103, 153)
-    );
+    --scrollBarThumbBackgroundColor: rgb(43, 101, 155);
     --scrollBarThumbBackgroundHoverColor: var(
         --colorSecondaryDarken1,
         rgb(58, 129, 191)
```
amalyah, line 360: Would suggest not referencing the RGB values in these definitions since we will be updating those values soon.

src/scss/common/_theme--dark.scss
```diff
@@ -354,23 +354,14 @@
...
     );
-    --scrollBarTrackBackgroundColor: var(
-        --colorTertiaryDarken1,
-        rgb(32, 50, 70)
-    );
-    --scrollBarTrackCornerBackgroundColor: var(
-        --colorTertiaryDarken1,
-        rgb(32, 50, 70)
-    );
+    --scrollBarTrackBackgroundColor: rgb(27, 45, 62);
```
amalyah, line 362: Please use new variable for the snowflakes/dark-surface design color.

src/scss/common/_theme--dark.scss
```diff
@@ -354,23 +354,14 @@
     ==========================================================================
     
   */
-
-    --scrollBarThumbBackgroundColor: var(
-        --colorSecondaryDarken2,
-        rgb(46, 103, 153)
-    );
+    --scrollBarThumbBackgroundColor: rgb(43, 101, 155);
```
amalyah, line 357: Use the --colorSecondaryDarken2 here to make the color changes easier later.

src/scss/common/_theme--light.scss
```diff
@@ -351,18 +351,11 @@
...
-        --colorQuaternaryDarken2,
-        #676a71
-    );
-    --scrollBarTrackBackgroundColor: transparent;
-    --scrollBarTrackCornerBackgroundColor: var(
-        --colorQuaternaryLighten4,
-        #f2f4f7
-    );
-
+    --scrollBarThumbBackgroundColor: var(--primary, rgb(0, 90, 143));
```
amalyah, line 354: In Sketch this is --primary at 70% opacity

.storybook/manager-head.html
```diff
@@ -1,159 +1,165 @@
...
+    .storybook-readme-story code.language-html,
+    .storybook-readme-story code.language-js {
+        color: #ccd5e0;
+    }
+
+    .storybook-readme-story .markdown-body table thead tr {
+        background-color: #141f2c;
+    }
+
+    /* scrollbars */
```
amalyah, line 114: Changes mentioned in other files would need to be applied here as well

src/components/rux-button/rux-button.js
```diff
@@ -1,247 +1,253 @@
...
+        return html`
+            <style>
+                :host {
+                    display: inline-flex;
+                }
 
-        .rux-button-group {
-          display: flex;
-        }
+                *[hidden] {
```
amalyah, line 38: Is there a reason that there are button changes in this PR?

amalyah (changes requested):

src/scss/core/select.scss
```diff
@@ -5,72 +5,73 @@
...
 .rux-select:focus::-ms-value {
-	background: transparent;
+    background: transparent;
 }
 
 .rux-select:invalid {
-  border: 1px solid var(--inputInvalidBorderColor);
-}
\ No newline at end of file
+    border: 1px solid var(--selectMenuInvalidBorder);
```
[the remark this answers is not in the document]
micahjones13: Definitely looks out of place. Time to figure out what 10 days ago Micah was thinking

src/scss/common/_variables.scss
```diff
@@ -1,220 +1,220 @@
...
+    --controlOptionSize: 1.125rem; /* 18px */
+    --controlBorderRadius: 3px;
+
+    --labelFontSize: 1rem;
+    --smallLabelFontSize: 0.875rem;
+    --largeLabelFontSize: 1.125rem;
+
+    /* Typography */
+    --fontFamily: 'Roboto', -apple-system, BlinkMacSystemFont, 'Segoe UI',
+        Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif;
```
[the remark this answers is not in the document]
micahjones13: The changes here are due to prettier, but I believe it's listed twice because after the first 'Roboto' the rest are lists of alternatives to use and are specific to certain platforms.

src/components/rux-button/rux-button.js
amalyah: Ah didn't realize that it would add new *[hidden] sections as well as change spacing and punctuation

micahjones13: It's really hard to tell what's happening in huge diffs like this, but it didn't add any new ones, it just re-writes it to be spaced correctly and that registers on github as a deletion and an addition.

src/scss/common/_global.scss
micahjones13: Screen Shot 2021-07-20 at 10 38 56 AM

Here's what it looks like with a 4px border-radius. Doesn't look rounded at all. This may be because of some of the other options like the `border-left-width` for the `::-webkit-scrollbar-thumb:vertical` but I'll need to look more into it.

micahjones13:
> for some reason, it looks like this breaks the select form component styles. can you take a look? its hard for me to see what's going on in the diff history

Just reverted the select scss to match current implementation in master - I'm not sure how this ended up here. I must have forgot to swap branches when I worked on some changes for select :no_good:

src/scss/core/select.scss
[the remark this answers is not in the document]
micahjones13: These have been reverted. I blame this on vacation brain.

src/scss/common/_global.scss
avillaindesire: Looking at the scrollbar, it’s built really weird. There’s a background-clip on the ::webkit-scrollbar-thumb which, when disabled, doubles the size of it. I think that's why you don't see the rounded corners with 4px. You need to double it to 8px to actually see it show up.

src/scss/common/_theme--light.scss
```diff
@@ -11,8 +11,9 @@
     --secondaryText: var(--colorQuaternaryDarken3, #51555b);
     --globalAppHeader: var(--colorTertiaryDarken2, #172635);
     --surfaceElements: var(--colorWhite, #ffffff);
-    --primary: var(--colorPrimary, #005a8f);
+    --primary: var(--colorPrimary, #035d91);
     --primaryLight: var(--colorPrimaryLighten1, #2f7aa7);
+    --primaryWithOpacity: rgba(var(--primary), 0.7);
```
NicDominguez, line 16: I had this same kind of change in my component and was unable to get the rgba function to compile properly from scss to css. The chrome dev tools is showing --scrollBarThumbBackgroundColor as undefined so I think you might have the same issue. I still need to talk to Sasha about this so maybe you can jump on a call with us. He said he knows a solution.

NicDominguez (approved):

src/scss/common/_theme--light.scss
```diff
@@ -13,6 +12,7 @@
     --surfaceElements: var(--colorWhite, #ffffff);
     --primary: var(--colorPrimary, #005a8f);
     --primaryLight: var(--colorPrimaryLighten1, #2f7aa7);
+    --primaryWithOpacity: rgba(0, 90, 143, 0.7);
```
micahjones13, line 15: Just went with this due to some limitations adding an opacity to a css variable. With the move to Stencil, I don't feel this is the largest priority. @Cheshire89 , I tried your scss variable way that you showed Nic and me, but couldn't get it to work. I think it has something to do with this being translated to CSS through gulp and having a SCSS var isn't valid CSS. Just a hunch, though. Thoughts?

src/scss/common/_theme--light.scss
```diff
@@ -363,18 +363,11 @@
...
-    );
-    --scrollBarTrackBackgroundColor: transparent;
-    --scrollBarTrackCornerBackgroundColor: var(
-        --colorQuaternaryLighten4,
-        #f2f4f7
-    );
-
+    --scrollBarThumbBackgroundColor: var(--primaryWithOpacity);
+    --scrollBarThumbBackgroundHoverColor: var(--colorPrimaryLighten1);
+    --scrollBarTrackBackgroundColor: var(--colorWhite, rgb(255, 255, 255));
```
amalyah, line 368: Nice to have: switch to surface element variable

amalyah (approved):
Approved pending Daniel's response to hover colors.

src/scss/common/_global.scss
avillaindesire: I think we talked about it before, but I'm not sure. In any case, I like it. Let's add it to the design files.

markacianfrani (approved):

PR #154, opened 2021-07-02
select element updated light/dark

micahjones13:
[JIRA Link](https://rocketcom.atlassian.net/browse/ASTRO-1529)

Updates the select element (rux-select) class to be in line with KM updates and current Abstract libraries. 
One notable change is adding a background color behind the SVG, which was accomplished by adding an additional param to `background-image` of a linear gradient, and then positioning that accordingly. 
The SVG color updates were left out of this description due to length. 
Abstract [Dark](https://app.abstract.com/projects/2d8cf270-0b93-11e8-89ff-61554354acb1/branches/master/files/1d17329d-e1a3-490a-a6ad-097e2ecd4825/pages/5DFC6D2A-252C-4612-B7B4-62409FB413A7?s=select%20menu) and [Light](https://app.abstract.com/projects/2d8cf270-0b93-11e8-89ff-61554354acb1/branches/master/files/b7c2bf0c-9dec-4b07-96d4-ead85393c9d0?s=multi-select) for reference. 
Light changes: 
```
--selectHoverOptBackgroundColor: rgb(28, 63, 94);
    --selectMenuBorderColor: rgb(100, 156, 189);
    --selectMenuCaretBackgroundColor: linear-gradient(
        rgba(206, 233, 252, 1),
        rgba(206, 233, 252, 1)
    );
    --selectMenuBackgroundColor: rgb(234, 238, 244);
    --selectMenuTextColor: var(--primary);
    --selectMenuInvalidBorder: var(--criticalBorder);
    --selectMenuTextHoverColor: rgb(0, 72, 114);
 
```
Dark changes: 
```
 --selectHoverOptBackgroundColor: rgb(28, 63, 94);
    --selectMenuBorderColor: rgb(43, 101, 155);
    --selectMenuCaretBackgroundColor: linear-gradient(
        rgba(28, 63, 94, 1),
        rgba(28, 63, 94, 1)
    );
    --selectMenuBackgroundColor: var(--backgroundColor);
    --selectMenuTextColor: var(--primary);
    --selectMenuInvalidBorder: var(--criticalBorder);
    --selectMenuTextHoverColor: rgb(146, 203, 255);

```

PR #157, opened 2021-07-09
Astro 1557 update slider km

micahjones13:
Updated the slider color vars and values to be in line with current abstract and KM standards. 
[Jira Link](https://rocketcom.atlassian.net/browse/ASTRO-1557)
Dark theme changes: 

1. `--sliderThumbBackgroundColor: var(--backgroundColor);` from `  --sliderThumbBackgroundColor: var(--primaryDark);`
2. `--sliderThumbBorderColor: var(--primary);` from `--sliderThumbBorderColor: var(--inputBackground);`
3. `--sliderTrackBackgroundColor: var(--primary);` from ` --sliderTrackBackgroundColor: var(--secondaryText);`
4. `    --sliderSelectedThumbBorderColor: var(--primaryLight);` from `--sliderSelectedThumbBorderColor: var(--primaryLight);`
5.  Upped `--sliderThumbBorderSize` to 2px instead of 1px, as per abstract 

Light theme: 

1. `--sliderThumbBackgroundColor: var(--backgroundColor);` from `--sliderThumbBackgroundColor: var(--primaryDark);`
2.  `--sliderThumbBorderColor: var(--primary);` from `--sliderThumbBorderColor: var(--inputBackground);`
3. `    --sliderTrackBackgroundColor: var(--primary);` from `--sliderTrackBackgroundColor: var(--inputBackground);`
4.  Upped `--sliderThumbBorderSize` to 2px instead of 1px, as per abstract 

The vars for hover values were not changed, I didn't see any slider hover files in abstract. If I'm just blind and didn't see them let me know. 
The massive file size is thanks to prettier.


src/components/rux-slider/rux-slider.js
```diff
@@ -1,395 +1,410 @@
...
-        }
-
-        .rux-range:disabled::-webkit-slider-thumb {
-          cursor: var(--disabledCursor);
-        }
-
-        .rux-range:not(:disabled)::-webkit-slider-thumb:active {
-          border-color: var(--sliderSelectedThumbBorderColor);
-          background-color: var(--inputBackground);
-          background: radial-gradient(circle,
```
micahjones13, line 259: Removed the radial gradient here, [current abstract](https://app.abstract.com/projects/2d8cf270-0b93-11e8-89ff-61554354acb1/branches/master/commits/26319f70445d912070d1e40bd7904fd95a4038c0/files/1d17329d-e1a3-490a-a6ad-097e2ecd4825/layers/7BC0B628-BEA4-41E8-90DF-2A15E49C8F64?mode=build) doesn't use it.

src/components/rux-slider/rux-slider.js
```diff
@@ -1,395 +1,410 @@
...
-            -webkit-user-select: none;
-            -khtml-user-select: none;
-              -moz-user-select: none;
-                -ms-user-select: none;
+    render() {
+        return html`
+            <style>
+                :host {
+                    --thumbSize: var(--controlOptionSize);
+                    --thumbShadow: 0 3px 5px rgba(0, 0, 0, 0.14),
```
amalyah, line 77: In the current designs, there are no shadows.

src/components/rux-slider/rux-slider.js
```diff
@@ -1,395 +1,410 @@
...
+                        0 1px 3px rgba(0, 0, 0, 0.2);
+                    --thumbShadowHover: 0 6px 10px rgba(0, 0, 0, 0.14),
+                        0 1px 18px rgba(0, 0, 0, 0.12),
+                        0 3px 5px rgba(0, 0, 0, 0.2);
+                    --thumbShadowActive: inset 0 0 0 4px var(--colorPrimary),
+                        0 1px 3px rgba(0, 0, 0, 0.14),
+                        0 1px 4px rgba(0, 0, 0, 0.12),
+                        0 1px 1px rgba(0, 0, 0, 0.2);
+
+                    --trackHeight: 3px;
```
amalyah, line 88: In the current designs, the unselected track is 1px high while the selected track is 5px high. (Relevant to multiple lines in this file.)

src/components/rux-slider/rux-slider.js
```diff
@@ -1,395 +1,410 @@
...
+                /* Track -> WebKit */
+                .rux-range::-webkit-slider-runnable-track {
+                    display: flex;
+                    align-items: center;
+                    max-width: 100%;
+
+                    /* width: 100%; */
+                    height: var(--trackHeight, 2px);
+                    cursor: var(--trackCursor, pointer);
+                    /* added */
```
amalyah, line 156: What does added mean here?

src/components/rux-slider/rux-slider.js
```diff
@@ -1,395 +1,410 @@
...
+                        var(--sliderThumbBorderColor);
+                    background-color: var(--sliderThumbBackgroundColor);
+
+                    cursor: pointer;
+                    box-shadow: inset 0 0 1px 0 rgba(255, 255, 255, 0.9),
+                        var(--thumbShadow);
+                }
+
+                .rux-range:not(:disabled)::-moz-range-thumb:active {
+                    background: radial-gradient(
```
amalyah, line 314: No radial gradient present in the current designs.

src/scss/common/_theme--dark.scss
```diff
@@ -114,18 +114,15 @@
     ==========================================================================
 
   */
-    --sliderThumbBackgroundColor: var(--primaryDark);
-    --sliderThumbBorderColor: var(--inputBackground);
+    --sliderThumbBackgroundColor: var(--backgroundColor);
+    --sliderThumbBorderColor: var(--primary);
     --sliderHoverThumbBackgroundColor: var(--primaryDark);
```
amalyah, line 119: --backgroundColor

src/scss/common/_theme--dark.scss
```diff
@@ -114,18 +114,15 @@
...
     --sliderHoverThumbBorderColor: var(--inputBackground);
```
amalyah, line 120: --primaryLight

src/scss/common/_theme--light.scss
```diff
@@ -129,18 +129,16 @@
     ==========================================================================
 
   */
-    --sliderThumbBackgroundColor: var(--primaryDark);
-    --sliderThumbBorderColor: var(--inputBackground);
+
+    --sliderThumbBackgroundColor: var(--backgroundColor);
+    --sliderThumbBorderColor: var(--primary);
     --sliderHoverThumbBackgroundColor: var(--primaryDark);
```
amalyah, line 135: --backgroundColor

src/scss/common/_theme--light.scss
```diff
@@ -129,18 +129,16 @@
...
     --sliderHoverThumbBorderColor: var(--inputBackground);
```
amalyah, line 136: --primaryLight
(@avillaindesire I suspect this is a design bug since usually hover in light goes to primaryDark. What do you think?)

src/scss/common/_theme--light.scss
```diff
@@ -129,18 +129,16 @@
...
-
     --sliderTrackBorderColor: var(--secondaryText);
```
amalyah, line 137: Is this in use right now? This is not a color we have anywhere in the current Slider designs.

src/scss/common/_theme--light.scss
```diff
@@ -129,18 +129,16 @@
...
-    --sliderTrackBackgroundColor: var(--inputBackground);
-
+    --sliderTrackBackgroundColor: var(--primary);
     --sliderSelectedThumbBorderColor: var(--primaryLight);
     --sliderSelectedTrackBackgroundColor: var(--primary);
-
-    --sliderThumbBorderSize: 1px;
+    --sliderThumbBorderSize: 2px;
     --sliderTrackBorderSize: 0.25px;
```
amalyah, line 142: 1px to match the dark theme

amalyah (changes requested):

amalyah:
Ah I see what happened. You can use the Pressed state in Abstract for Hover. This brings up some need for better naming or at least commenting functions in the design files. We will work on that in the future. @avillaindesire do you agree with using the Pressed state for the hover state as well?

avillaindesire:
> Ah I see what happened. You can use the Pressed state in Abstract for Hover. This brings up some need for better naming or at least commenting functions in the design files. We will work on that in the future. @avillaindesire do you agree with using the Pressed state for the hover state as well?

Yep, we can use the pressed state for hover. In the future, we may want to explore a different state for hover since you can select any point on the track to move the thumb there.

src/scss/common/_theme--light.scss
avillaindesire: Yep, it looks like a design bug. It should go to --primaryDark

NicDominguez (approved):
With the design changes requested by Amalya and Daniel this looks good. The merge conflicts are probably only a couple corrected rgb values so won't mess anything up.

avillaindesire (changes requested):
The slider was was one of the components that required more than a color change. As @amalyah pointed out, the left side of the thumb should be 5px high and the left 1px. The pressed state should not change the fill (shouldn't go primary blue with a white dot). The disabled state has a weird thing where it is making all of the elements transparent. This means that you can see the track behind the thumb. We shouldn't be seeing that. let's make sure that the thumb is obscuring the track even in the disabled state.

src/scss/common/_theme--light.scss
amalyah: Ticket for design fix: https://rocketcom.atlassian.net/browse/ASTRO-1585

markacianfrani (changes requested):
the select styles are broken here. do we need to merge https://github.com/RocketCommunicationsInc/astro-components/pull/154 in first?

src/scss/common/_theme--light.scss
micahjones13: It's not, I'll remove it

src/components/rux-slider/rux-slider.js
amalyah: Is there a reason to keep these in @micahjones13?

src/components/rux-slider/rux-slider.js
```diff
@@ -178,14 +179,24 @@ export class RuxSlider extends LitElement {
                 .rux-range::-moz-range-track {
                     display: flex;
                     align-items: center;
-
+                    max-width: 100%;
                     /* width: 100%; */
-                    height: var(--trackHeight);
                     cursor: var(--trackCursor, pointer);
+                    border-radius: 2.5px;
+                    min-height: var(--trackHeight);
```
amalyah, line 186: Is there a reason that we're switching from fixed height to min- and max-heights?

src/scss/common/_theme--dark.scss
```diff
@@ -120,18 +120,15 @@
...
+    --sliderThumbBorderColor: var(--primary);
+    --sliderHoverThumbBackgroundColor: var(--backgroundColor);
+    --sliderHoverThumbBorderColor: var(--primaryLight);
+    --sliderTrackBackgroundColor: var(--primary);
     --sliderSelectedThumbBorderColor: var(--primaryLight);
     --sliderSelectedTrackBackgroundColor: var(--primary);
-
-    --sliderThumbBorderSize: 1px;
+    --sliderThumbBorderSize: 2px;
+    --sliderThumbSize: 19px;
```
amalyah, line 131: Nice to have: The slider thumb size, slider border size, and slider track border size are the same between themes. Would suggest keeping that in a slider-specific file instead of a theme-specific file.

src/scss/common/_theme--light.scss
```diff
@@ -131,18 +131,16 @@
...
+    --sliderHoverThumbBackgroundColor: var(--backgroundColor);
+    --sliderHoverThumbBorderColor: var(--primaryLight);
+    --sliderTrackBackgroundColor: var(--primary);
     --sliderSelectedThumbBorderColor: var(--primaryLight);
     --sliderSelectedTrackBackgroundColor: var(--primary);
-
-    --sliderThumbBorderSize: 1px;
+    --sliderThumbBorderSize: 2px;
+    --sliderThumbSize: 19px;
     --sliderTrackBorderSize: 0.25px;
```
amalyah, line 144: I suspect that this is supposed to be the same value in both themes (right now they're different).

amalyah (changes requested):

src/components/rux-slider/rux-slider.js
micahjones13: Sorry, an oversight on my part. Shadows have been removed.

src/components/rux-slider/rux-slider.js
micahjones13: Yes - because we're using a native HTML input[type=range], it's really difficult to style the height of each side of the thumb differently, hence the linear-gradient approach I took. If we set height explicitly here, then all of the slider track will be that height and no more or less, thus making the different heights impossible. This way, the background-size is being set depending on which end of the thumb it's on, and that gives us the 5px and the 1px height. It can't be less than 1px or more than 5px though, hence max/min-height

src/scss/common/_theme--dark.scss
micahjones13: Great point. Also removing --sliderTrackBorderSize because it's not being used

amalyah (approved):

markacianfrani (approved):

PR #158, opened 2021-07-12
Astro 1559 km radio button update

NicDominguez:
Updates radio button dark-theme, light-theme, and core scss for KM updates.
[JIRA Ticket](https://rocketcom.atlassian.net/browse/ASTRO-1559)

Dark Theme Changes:
`--controlBorderColor: var(--primary) >>>  var(--colorSecondaryDarken2)`
`--controlSelectedBorderColor: var(--primary) >>> var(--colorSecondaryDarken2);`

Light Theme Changes:
`--controlBorderColor: var(--primary) >>> var(--colorPrimaryLighten2)`

Core Changes:
```
.rux-radio-button input[type="radio"]:checked + label::before {
    border-color: var(--controlSelectedOutlineBorderColor) >>> var(--controlSelectedBorderColor);;
}
```


markacianfrani:
should this also be affecting the styles of checkbox and switch?

NicDominguez:
> should this also be affecting the styles of checkbox and switch?

Yes, they share the same global variables and all components have been are in the process of being updated to the new values just haven't been merge yet.

src/scss/core/radio.scss
```diff
@@ -17,80 +17,84 @@
...
+    margin: 0 0.625rem 0 0;
+    border: 1px solid var(--controlBorderColor);
 }
 
-.rux-radio-button input[type="radio"] + label::before {
-  border-radius: 100%;
-  background-color: transparent;
+.rux-radio-button input[type='radio'] + label::before {
+    border-radius: 100%;
+    background-color: transparent;
```
amalyah, line 59: Not sure if this is the relevant line, but the radio buttons no longer have transparent backgrounds. They use --backgroundColor now instead (for both themes).

src/scss/common/_theme--light.scss
```diff
@@ -81,16 +81,17 @@
     --controlTextColor: var(--primaryElementText);
     --controlLabelColor: var(--defaultText);
     --controlBackgroundColor: var(--primary);
-    --controlBorderColor: var(--primary);
+    --controlBorderColor: var(--colorPrimaryLighten2);
     --controlAccentColor: var(--primary);
 
     --controlSelectedTextColor: var(--defaultText);
     --controlSelectedBackgroundColor: transparent;
```
amalyah, line 88: This might be out of scope but it seems weird that this color is transparent in light theme but has a value in dark theme.

src/scss/common/_theme--light.scss
```diff
@@ -81,16 +81,17 @@
...
     --controlSelectedBorderColor: var(--primary);
```
amalyah, line 89: --colorPrimaryLighten2

src/scss/common/_theme--light.scss
```diff
@@ -81,16 +81,17 @@
...
-
-    --controlHoverOutlineBackgroundColor: none;
     --controlHoverBorderColor: var(--primaryDark);
+
     --controlSelectedOutlineBorderColor: var(--primary);
+
+    --controlHoverOutlineBackgroundColor: none;
```
amalyah, line 94: Out of scope, but it seems odd that the dark theme file doesn't have this variable or the variable on line 95.

amalyah (changes requested):

src/scss/common/_theme--light.scss
NicDominguez: I've changed it to the value matching what would normally show with transparent. Visual outcome is the same.

src/scss/common/_theme--light.scss
NicDominguez: I've added them to dark theme to keep the files the same but `--controlHoverOutlineBackgroundColor` appears to have no use and must be left over from a previous design.

src/scss/core/radio.scss
```diff
@@ -56,7 +56,6 @@
 
 .rux-radio-button input[type='radio'] + label::before {
     border-radius: 100%;
-    background-color: transparent;
```
amalyah: Is there a declaration elsewhere in the file that states that the background-color is always --backgroundColor (or some variable that ties up to that eventually)?

src/scss/common/_theme--light.scss
amalyah: Ah okay. Thanks!

src/scss/core/radio.scss
NicDominguez: The code relied on the default value for background-color being transparent. I have added new variables to explicitly declare the background color which should make it easier to update in the future.

amalyah (approved):

Cheshire89 (approved):

markacianfrani (approved):

PR #161, opened 2021-07-14
Astro 1558 checkbox km

Cheshire89:
https://rocketcom.atlassian.net/browse/ASTRO-1558

- Updated dev storybook build to run gulp in parallel
- removed __variables.scss and associated gulp tasks as tests with Nick showed that this file is no longer used anywhere
- added Primary, Secondary, Tertiary, and Quaternary colors from figma color sheet to variables file
- updated checkbox colors as per designs using newly added variables

* new colors are scoped to checkbox and radio buttons as not to break rest of the app


src/scss/common/_theme--dark.scss
```diff
@@ -81,17 +81,15 @@
     For checkboxes, radio buttons etc …
```
amalyah, line 81: Nice to Have: As a note, it would be a lot easier for me to do reviews on these controls if they actually had separate variables so that the variable names could make more sense in context. Just from this file I can't tell which colors are just used in some controls and not others or what "...TextColor" variables are referring to for each control.

src/scss/common/_theme--light.scss
```diff
@@ -78,18 +78,15 @@
     For checkboxes, radio buttons etc …
 
   */
-    --controlTextColor: var(--primaryElementText);
+    --controlTextColor: var(--colorPrimary500);
```
amalyah, line 81: Nice to have: --primary

src/scss/common/_theme--light.scss
```diff
@@ -78,18 +78,15 @@
...
     --controlLabelColor: var(--defaultText);
     --controlBackgroundColor: var(--primary);
-    --controlBorderColor: var(--primary);
-    --controlAccentColor: var(--primary);
+    --controlBorderColor: var(--colorPrimary600);
```
amalyah, line 84: Just FYI, I think another PR is using --primaryLighter as an alias for this.

amalyah, line 84: Also, I believe that this is supposed to be the equivalent of --colorPrimary300

src/scss/common/_theme--light.scss
```diff
@@ -78,18 +78,15 @@
...
 
     --controlSelectedTextColor: var(--defaultText);
-    --controlSelectedBackgroundColor: transparent;
-    --controlSelectedBorderColor: var(--primary);
+    --controlSelectedBackgroundColor: var(--colorQuaternary200);
```
amalyah, line 87: Nice to have: --backgroundColor

src/scss/common/_theme--light.scss
```diff
@@ -78,18 +78,15 @@
...
 
-    --controlHoverOutlineBackgroundColor: none;
-    --controlHoverBorderColor: var(--primaryDark);
+    --controlHoverBorderColor: var(--colorPrimary300);
```
amalyah, line 89: Nice to have: --primaryDark

src/scss/common/_theme--light.scss
```diff
@@ -78,18 +78,15 @@
...
     --controlSelectedOutlineBorderColor: var(--primary);
```
amalyah, line 90: Not sure what this control variable is used for, but it doesn't match the pattern in the dark theme. My guess is that this is supposed to be the same as --controlBorderColor's variable.

src/scss/common/_theme--light.scss
```diff
@@ -78,18 +78,15 @@
...
     --controlSelectedOutlineBackgroundColor: none;
```
amalyah, line 91: This does not seem to have an equivalent in the dark theme file. Seems odd.

amalyah (changes requested):

src/scss/core/table.scss
```diff
@@ -91,5 +91,5 @@
 }
 .rux-table tr:hover .rux-checkbox input[type="checkbox"]:not(:checked) + label::before,
 .rux-table tr:hover .rux-checkbox input[type="radio"] + label::before {
-	border-color: var(--controlSelectedBorderColor);
+	border-color: var(--primary);
```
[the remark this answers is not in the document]
Cheshire89: controlSelectedBorderColor was set to primary
Since control* variables are reserved for checkboxes and radio buttons I've substituted the value as not to cause a confusion down the line

src/scss/common/_variables.scss
```diff
@@ -222,4 +222,44 @@
     --colorSnowflakesLightShadow: #828181;
     --colorSnowflakesDarkSelected: #1c3f5e;
     --colorSnowflakesLightSelected: #cee9fc;
-}
+
+    --colorPrimary100: #CBDEE9;
```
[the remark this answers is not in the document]
Cheshire89: Correct and because we are doing incremental changes we cant yet override colors such as primaryDark and primary because it will override the entire app

markacianfrani:
"Updated dev storybook build to run gulp in parallel"

Can you explain what problem this solves?

Cheshire89:
> "Updated dev storybook build to run gulp in parallel"
> 
> Can you explain what problem this solves?

before when you ran  `npm run storybook` it would only recompile if html templates changes and would not recognize changes happening to scss files. One would have to open second console window and run gulp in order for global styles to be compiled. Updated command get rid of the need for the second terminal window and runs storybook build and scss compilation in parallel like we have in stencil repo

src/scss/common/_variables.scss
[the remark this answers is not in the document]
NicDominguez: I think we may want to wait to insert these and do them as a separate ticket. Then we can change the naming convention and make sure the regression still passes. I am totally on board with doing this change though as the 100-900 convention makes a lot more sense to me.

src/scss/common/_theme--dark.scss
```diff
@@ -81,17 +81,15 @@
...
 
   */
-    --controlTextColor: var(--primaryElementText);
+    --controlTextColor: var(--colorSecondary500);
```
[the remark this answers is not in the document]
NicDominguez: Just for reference, all color[...]500 colors are equivalent to the base color variable color[...]. So --colorSecondary500 in the new convention is equivalent to --colorSecondary in the current convention.

NicDominguez (changes requested):
Since the radio button change has yet to be merged we have some differences in component level variables.  The additions I made should only be used by the radio buttons but you may want to reference them if it makes sense in checkbox. This is what my variables in the radio button pr currently look like:

Dark Theme - 
```    
    --controlTextColor: var(--primaryElementText);
    --controlLabelColor: var(--defaultText);
    --controlBackgroundColor: var(--primary);
    --controlOutlineBackgroundColor: var(--backgroundColor);
    --controlBorderColor: var(--colorSecondaryDarken2);
    --controlAccentColor: var(--primary);

    --controlSelectedTextColor: var(--defaultText);
    --controlSelectedBackgroundColor: var(--primaryDark);
    --controlSelectedBorderColor: var(--colorSecondaryDarken2);
    --controlHoverBorderColor: var(--primaryLight);

    --controlSelectedOutlineBorderColor: var(--primary);

    --controlHoverOutlineBackgroundColor: var(--backgroundColor);
    --controlSelectedOutlineBackgroundColor: var(--backgroundColor); 
``` 

Light Theme - 
```
    --controlTextColor: var(--primaryElementText);
    --controlLabelColor: var(--defaultText);
    --controlBackgroundColor: var(--primary);
    --controlOutlineBackgroundColor: var(--backgroundColor);
    --controlBorderColor: var(--colorPrimaryLighten2);
    --controlAccentColor: var(--primary);

    --controlSelectedTextColor: var(--defaultText);
    --controlSelectedBackgroundColor: var(--primary);
    --controlSelectedBorderColor: var(--colorPrimaryLighten2);
    --controlHoverBorderColor: var(--primaryDark);

    --controlSelectedOutlineBorderColor: var(--primary);

    --controlHoverOutlineBackgroundColor: var(--backgroundColor);
    --controlSelectedOutlineBackgroundColor: var(--backgroundColor);
```

NicDominguez (approved):

markacianfrani (approved):

amalyah (approved):

src/scss/core/radio.scss
```diff
@@ -17,80 +17,84 @@
...
+    + label:before {
+    border-color: var(--controlHoverBorderColor);
 }
 
-.rux-radio-button input[type="radio"]:not(:disabled):checked:hover + label:after{
-  background-color: var(--controlHoverBorderColor);
+.rux-radio-button
+    input[type='radio']:not(:disabled):checked:hover
+    + label:after {
+    background-color: var(--controlBackgroundColor);
```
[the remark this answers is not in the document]
Cheshire89: radio and checkbox dont have a transparent background (design does have a background set)

Issue #186, opened 2022-01-31
3 Digit Julian Date


Rachel-Platz:
I would like to view the Julian Date in the rux-clock so that it always shows 3 digits. Attached is a screenshot of the the current implementation from the astro storyboard where the clock's date only has 2 digits when it is less than 100. There is also a screenshot from juliandate.net to show that the day always has 3 digits and a link to IBM that shows that the Julian date format should be 3 digits.

IBM Link:
https://www.ibm.com/docs/en/cics-vt/2.1?topic=reengineering-converting-between-julian-gregorian-date-formats

![AstroStorybook RuxClock](https://user-images.githubusercontent.com/98778932/151872915-076ef5b3-918c-4449-adf8-380ec868c5f9.PNG)
![Juliandate](https://user-images.githubusercontent.com/98778932/151872927-b8b84d06-cd70-45a7-8042-17a9648b5558.PNG)

dmcalester:
Thanks Rachel,

We have an internal ticket and adding this to our sprint that started today.

dmcalester:
Resolved in 6.1.0 release https://github.com/RocketCommunicationsInc/astro/releases/tag/v6.1.0

PR #187, opened 2022-02-01
fix(clock): julien date is now always 3 digits

micahjones13:
Updates `${this.dayOfYear}` to `${this.dayOfYear.toString().padStart(3, '0')}` to ensure that the date is always 3 digits with leading zero's if necessary. 

Issue: https://github.com/RocketCommunicationsInc/astro-components/issues/186


markacianfrani (approved):

micahjones13:
@markacianfrani can you merge this? It needs the status checks to pass first but I'm pretty sure they never will since this isn't linked to SB anymore.

